PR #15, opened 2020-04-05
Added an option to override the decorator generator

EricRabil:
Hi, in this PR I simply added a constructor parameter that allows one to override the decorator function.

I found the lack of customization to be inconvenient, and this PR solves the problem I was encountering.

### DIFF
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -73,17 +73,17 @@ export const createDecorations = (state, awareness) => {
  * @param {Awareness} awareness
  * @return {any}
  */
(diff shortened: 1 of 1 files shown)


EricRabil:
I updated the PR to override the portion of the `createDecorations` function responsible for composing the `HTMLElement`. This standardizes the positioning of the cursor while allowing its composition to be overridden.

amilich:
This looks super helpful

dmonad:
Hey @EricRabil , thanks for this feature! Sorry it took so long. I change the implementation slightly, so that yCursorPlugin accepts an options object instead of a cursorBuilder parameter. I also documented this addition in the README.

PR #18, opened 2020-05-19
Unexpected scroll in the prose mirror view by yjs update

boschDev:
What Happens:
1. There are 2 different clients connected to the same room (ClientA, ClientB) and they both have the cursor on the same prose mirror view (and thus both are focused on the view).
2. ClientA sets the cursor somewhere in the text field and scrolls up or down until the cursor is outside of the view.
3. ClientB types something in the same text field (it does not matter where).

Expected behaviour:
The view of ClientA should not scroll because the cursor of this client is not visible in the browser. 

Actual behaviour:
The view of ClientA will be scrolled back to where the client previously placed cursor.

Note: The issue is only reproducable with 2 different computers. If not the prose mirror view of one of the clients will lose focus when going the the other browser window.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -13,6 +13,7 @@ import * as error from 'lib0/error.js'
 import * as Y from 'yjs'
 import { absolutePositionToRelativePosition, relativePositionToAbsolutePosition } from '../lib.js'
 import * as random from 'lib0/random.js'
+import * as environment from 'lib0/environment.js'
 
 /**
  * @param {Y.Item} item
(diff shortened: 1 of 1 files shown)


src/plugins/sync-plugin.js
```diff
@@ -178,6 +180,39 @@ export const getRelativeSelection = (pmbinding, state) => ({
...
 
+const getElementFromTextNode = node => {
+  if (dom.checkNodeType(node, dom.ELEMENT_NODE)) {
+    return node
+  }
+  return node.parentElement
+}
+
+const isDomSelectionInView = () => {
+  const selection = window.getSelection()
```
dmonad, line 191: This should be `view._root.getSelection()` so it works properly in web components (they have a different document root).

src/plugins/sync-plugin.js
```diff
@@ -178,6 +180,39 @@ export const getRelativeSelection = (pmbinding, state) => ({
...
+  const anchorElement = getElementFromTextNode(selection.anchorNode)
```
dmonad, line 192: I think it would be easier to just compute the bounding rect of the selection range directly. This approach is more granular and will also account for the case that there is a huge paragraph with a selection:

```js
// const range = selection.getRangeAt(0) // this is an alternative approach to get the range, but it is marked as experimental on mdn
// This is why we create a selection range manually:
const range = view._root.createRange()
range.setStart(selection.anchorNode, selection.anchorOffset)
range.setEnd(selection.focusNode, selection.focusOffset)
const rect = range.getBoundingClientRect()
```

src/plugins/sync-plugin.js
boschDev: That is a good one. Didn't think about the web components. Moved this function also to the ProsemirrorBinding class.

boschDev:
The `npm run lint` now runs without errors on my pc. Also thanks for the good suggestions. Those are also implemented

boschDev:
@dmonad any updates on this pr?

rebolyte:
Ran into this behavior as well. Any chance we can get this merged @dmonad?

PR #19, opened 2020-05-21
Cusrsors with multiple prose mirror views in one yjs doc and awearness

boschDev:
This is a optional parameter for the cursor plugin.
First I wanted to make an id of the _item.id to combine the client and the id. But when I checked it on a yXmlFragment on the root of the yDoc it seems that the _item was null. Maybe is missed something to get a unique id for each yXmlFragments.
The first implementation I tried all the cursors would be saved in a different key name in the awareness state. After the first test that was not a great solution. If you have a lot of prose mirror views there will also be a lot of keys in that awareness state. I kept this in when somebody wants to change the cursor state name.
After this I added an id in the cursor info and only rendered the cursor when the id is correct for that prose mirror view. I will also open a pr in the [yjs demo's](https://github.com/yjs/yjs-demos) where I changed the prose mirror example to 2 prose mirror views with cursors. That was also the case I used to test this.

### DIFF
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -32,14 +32,13 @@ export const defaultCursorBuilder = user => {
 }
 
 /**
- * @param {string} cursorStateName
  * @param {string} cursorId
  * @param {any} state
  * @param {Awareness} awareness
  * @param {Function} createCursor
  * @return {any} DecorationSet
  */
-export const createDecorations = (cursorStateName, cursorId, state, awareness, createCursor) => {
+export const createDecorations = (cursorId, state, awareness, createCursor) => {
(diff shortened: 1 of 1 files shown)


anilgulecha:
@boschDev is this usable? I too have a scenario where there's multiple tiptaps backed by single yjs doc and this PR can help

dmonad:
I assume the issue was that the cursor position jumped around because both instances wanted to control the position.

This was a bug. This should always have been possible without giving a name to the prosemirror binding. All editor bindings use the 'cursor' field to propagate cursor information. I fixed the bug and also specified a field `'cursorStateField'` that allows you to specify a different cursor field (in case you want to show two selections from two editors).

I published a new release 0.3.7 that fixes the issue.

PR #26, opened 2020-10-13
feat: Add utility methods for conversion to/from Prosemirror

tommoor:
As discussed here: https://discuss.yjs.dev/t/prosemirror-serializer/240

This PR adds new methods to aid getting content into and out of Y.Doc format for Prosemirror users. As part of this I had to move some more generic methods out of the `syncPlugin` and up to the `lib`, I also pulled the plugin keys to a higher level to avoid circular dependencies. 

`yDocToProsemirrorJSON` Is probably not _perfect_ yet, there might be an edge case I haven't covered, but that should become readily apparent in production usage soon.

Edit: I've published a fork with these utilities `@tommoor/y-prosemirror`

### DIFF
--- a/src/interface.js
+++ b/src/interface.js
@@ -120,5 +120,8 @@ export function yDocToProsemirrorJSON (
     return response
   }
 
-  return items.map(serialize)
+  return {
+    type: 'doc',
+    content: items.map(serialize)
+  }
 }
--- a/src/interface.js
+++ b/src/interface.js
@@ -1,0 +1,124 @@
+// eslint-disable-next-line
+import { Node, Schema } from 'prosemirror-model'
+import flatten from 'lodash.flatten'
+import { updateYFragment } from './lib'
+import * as Y from 'yjs'
+
+/**
+ * Utility method to convert a Prosemirror Doc Node into a Y.Doc.
+ *
+ * This can be used when importing existing content to Y.Doc for the first time,
+ * note that this should not be used to rehydrate a Y.Doc from a database once
+ * collaboration has begun as all history will be lost
+ *
+ * @param {Node} doc
+ * @param {string} xmlFragment
+ * @return {Y.Doc}
+ */
+export function prosemirrorToYDoc (doc, xmlFragment = 'prosemirror') {
+  const ydoc = new Y.Doc()
+  const type = ydoc.get(xmlFragment, Y.XmlFragment)
+  if (!type.doc) {
+    return ydoc
+  }
+
+  return updateYFragment(type.doc, type, doc, new Map())
+}
+
+/**
+ * Utility method to convert Prosemirror compatible JSON into a Y.Doc.
+ *
+ * This can be used when importing existing content to Y.Doc for the first time,
+ * note that this should not be used to rehydrate a Y.Doc from a database once
+ * collaboration has begun as all history will be lost
+ *
+ * @param {Schema} schema
+ * @param {any} state
+ * @param {string} xmlFragment
+ * @return {Y.Doc}
+ */
+export function prosemirrorJSONToYDoc (schema, state, xmlFragment = 'prosemirror') {
+  const doc = Node.fromJSON(schema, state)
+  return prosemirrorToYDoc(schema, doc, xmlFragment)
+}
+
+/**
+ * Utility method to convert a Y.Doc to a Prosemirror Doc node.
+ *
+ * @param {Schema} schema
+ * @param {Y.Doc} ydoc
+ * @return {Node}
+ */
+export function yDocToProsemirror (schema, ydoc) {
+  const state = yDocToProsemirrorJSON(ydoc)
+  return Node.fromJSON(schema, state)
+}
+
+/**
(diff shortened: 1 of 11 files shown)


src/interface.js
```diff
@@ -0,0 +1,124 @@
...
+ * @return {Y.Doc}
+ */
+export function prosemirrorToYDoc (doc, xmlFragment = 'prosemirror') {
+  const ydoc = new Y.Doc()
+  const type = ydoc.get(xmlFragment, Y.XmlFragment)
+  if (!type.doc) {
+    return ydoc
+  }
+
+  return updateYFragment(type.doc, type, doc, new Map())
```
tommoor, line 25: Perhaps the `mapping` argument for `updateYFragment` should be made optional?

src/interface.js
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * @type {Object} NodeObject
+     * @property {string} NodeObject.type
+     * @property {Record<string, string>=} NodeObject.attrs
+     * @property {Array<NodeObject>=} NodeObject.content
+     */
+    let response
+
+    // TODO: Must be a better way to detect text nodes than this
+    if (!item.nodeName) {
```
tommoor, line 80: Open to suggestions here

src/lib.js
```diff
@@ -168,3 +180,445 @@ export const relativePositionToAbsolutePosition = (y, documentType, relPos, mapp
...
+    if (left + right < pChildCnt) {
+      const ins = []
+      for (let i = left; i < pChildCnt - right; i++) {
+        ins.push(createTypeFromTextOrElementNode(pChildren[i], mapping))
+      }
+      yDomFragment.insert(left, ins)
+    }
+  }, ySyncPluginKey)
+
+  return y
```
tommoor, line 616: Note: `updateYFragment` now returns the `Y.Doc`, it didn't before

src/lib.js
```diff
@@ -3,9 +3,21 @@ import { ProsemirrorMapping } from './plugins/sync-plugin.js' // eslint-disable-
...
+import * as math from 'lib0/math.js'
+import * as object from 'lib0/object.js'
+import { simpleDiff } from 'lib0/diff.js'
+import { ySyncPluginKey } from './plugins/keys.js'
+
+/**
+ * @param {Y.Item} item
+ * @param {Y.Snapshot} [snapshot]
+ */
+export const isVisible = (item, snapshot) => snapshot === undefined ? !item.deleted : (snapshot.sv.has(item.id.client) && /** @type {number} */ (snapshot.sv.get(item.id.client)) > item.id.clock && !Y.isDeleted(snapshot.ds, item.id))
```
tommoor, line 20: These methods are all just moved from `syncPlugin` without edits, apart from `updateYFragment`, as noted below.

src/plugins/undo-plugin.js
```diff
@@ -31,10 +30,10 @@ export const yUndoPlugin = ({ protectedNodes = new Set(['paragraph']), trackedOr
       const undoManager = new UndoManager(ystate.type, {
         trackedOrigins: new Set([ySyncPluginKey].concat(trackedOrigins)),
         deleteFilter: item => !(item instanceof Item) ||
-                              !(item.content instanceof ContentType) ||
-                              !(item.content.type instanceof Text ||
-                              (item.content.type instanceof XmlElement && protectedNodes.has(item.content.type.nodeName))) ||
-                              item.content.type._length === 0
+          !(item.content instanceof ContentType) ||
```
tommoor, line 33: These formatting changes are the result of running `standard --fix`

tommoor:
I'd like to add some tests but the current test suite seems to be already failing on master…

dmonad:
Hi @tommoor 

thanks for adding this feature!

It takes some time for me to review a PR this huge. The refactor didn't make sense to me as I believe the ydoc⇔prosemirror binding internals shoulds live in `sync-plugin.js`. I'm very well aware that I have circular dependencies. Some bundlers throw warnings, but they are well supported by bundlers and they are actually supported by the spec itself. There are some caveats when working with circular dependencies, which is why you should avoid them if possible. But I'm not going to sacrifice a sound project structure to suppress the warning of circular dependencies. 

I reverted the refactoring & the formatting changes, and added some tests. The actual change is much simpler for me to comprehend: https://github.com/yjs/y-prosemirror/pull/29

This project is used by quite a lot of projects and I can't have it fail in production because I wasn't able to review the whole refactor. The next time maybe you could add fewer changes.

This is published in y-prosemirror@1.0.0 (it was time for a stable release)

tommoor:
> I reverted the refactoring & the formatting changes, and added some tests. The actual change is much simpler for me to comprehend

Yep, fair – just happy to have this merged. My logic at the time was that it felt strange to be using method from the y-sync plugin file inside of utils, and yes causes import loops – in hindsight it makes sense to reduce the diff.

dmonad:
No worries, you are not the first one who wanted to refactor one of my projects. I can't count how many PRs I rejected that added semicolons everywhere ^^

This is a helpful PR so thanks. Just saying this for the next one ;)

PR #33, opened 2020-12-09
fix: Scrolling behavior when remote clients are editing

### LINKED ISSUE #31: Scroll position changes when other clients are editing
### To Reproduce

- Create a doc long enough to scroll
- In one client place your cursor at the bottom of the screen and scroll to the top (common behavior when reading)
- In another client type at the top of the doc
- The scroll position in the first client will jump to the bottom

### Screencast

From the `y-prosemirror` demo site:

Kapture 2020-12-08 at 15 01 08

I tracked it down to this line – could you enlighten in which circumstances this is supposed to trigger? It feels like it would be fine if remote edits never caused the local scroll state to change.

https://github.com/yjs/y-prosemirror/blob/52116dd0333fa722b060b75ed86fb81f361b9e9a/src/plugins/sync-plugin.js#L344-L346

tommoor:
Only consider the dom selection if it's within the editor bounds.

closes #31

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -228,16 +228,24 @@ export class ProsemirrorBinding {
   }
 
   _isDomSelectionInView () {
-    // We only want to consider a dom selection that's within the editor itself
-    const documentElement = dom.doc.documentElement
-    if (!this.prosemirrorView.dom.contains(documentElement)) return false
-
     const selection = this.prosemirrorView._root.getSelection()
 
     const range = this.prosemirrorView._root.createRange()
     range.setStart(selection.anchorNode, selection.anchorOffset)
     range.setEnd(selection.focusNode, selection.focusOffset)
 
+    // This is a workaround for an edgecase where getBoundingClientRect will
+    // return zero values if the selection is collapsed at the start of a newline
+    // see reference here: https://stackoverflow.com/a/59780954
+    const rects = range.getClientRects()
+    if (!rects.length) {
+      // probably buggy newline behavior, explicitly select the node contents
+      if (range.startContainer && range.collapsed) {
+        range.selectNodeContents(range.startContainer)
+      }
+    }
+
+    const documentElement = dom.doc.documentElement
     const bounding = range.getBoundingClientRect()
 
     return bounding.bottom >= 0 && bounding.right >= 0 &&
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -218,7 +218,7 @@ export class ProsemirrorBinding {
   _isLocalCursorInView () {
     if (!this.prosemirrorView.hasFocus()) return false
     if (environment.isBrowser && this._domSelectionInView === null) {
-      // Calculte the domSelectionInView and clear by next tick after all events are finished
+      // Calculate the domSelectionInView and clear by next tick after all events are finished
       setTimeout(() => {
         this._domSelectionInView = null
       }, 0)
@@ -228,6 +228,10 @@ export class ProsemirrorBinding {
   }
 
   _isDomSelectionInView () {
+    // We only want to consider a dom selection that's within the editor itself
+    const documentElement = dom.doc.documentElement
+    if (!this.prosemirrorView.dom.contains(documentElement)) return false
+
     const selection = this.prosemirrorView._root.getSelection()
 
     const range = this.prosemirrorView._root.createRange()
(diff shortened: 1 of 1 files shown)


src/plugins/sync-plugin.js
```diff
@@ -252,7 +255,7 @@ export class ProsemirrorBinding {
   unrenderSnapshot () {
     this.mapping = new Map()
     this.mux(() => {
-      const fragmentContent = this.type.toArray().map(t => createNodeFromYElement(/** @type {Y.XmlElement} */ (t), this.prosemirrorView.state.schema, this.mapping)).filter(n => n !== null)
+      const fragmentContent = this.type.toArray().map(t => createNodeFromYElement(/** @type {Y.XmlElement} */(t), this.prosemirrorView.state.schema, this.mapping)).filter(n => n !== null)
```
tommoor, line 258: All of the edits from here down are a result of running `standard --fix` before committing vvv

src/plugins/sync-plugin.js
```diff
@@ -228,14 +228,17 @@ export class ProsemirrorBinding {
   }
 
   _isDomSelectionInView () {
+    // We only want to consider a dom selection that's within the editor itself
+    const documentElement = dom.doc.documentElement
+    if (!this.prosemirrorView.dom.contains(documentElement)) return false
```
tommoor, line 233: ref: https://developer.mozilla.org/en-US/docs/Web/API/Node/contains#Browser_compatibility

src/plugins/sync-plugin.js
dmonad: I care about these things more than I should. I use the space consistently between projects, so I need to reverse your changes every time I merge your PRs.. (small price because your PRs are much appreciated)

The package wouldn't publish if that was not validly formatted. I use standard v16 (the current version). This package uses v12, which also doesn't define that rule.

For the future, you could just run `npm run lint` using the intended version.

src/plugins/sync-plugin.js
dmonad: That doesn't make sense to me.. How can the `document` be inside the proseMirrorView? This doesn't check selections at all.

So is this the only actual change? Is the rest just linting?

src/plugins/sync-plugin.js
tommoor: >  I need to reverse your changes every time I merge your PRs..

I'm not _trying_ to change anything, I'm just running the included linters. It would be very useful if you'd consider setting up CI on these projects, this is the only project I contribute to that doesn't enforce linting rules automatically, it would save your time in the long run I'm sure.

src/plugins/sync-plugin.js
tommoor: It seems I misunderstood, will take another bash at it in the next few hours

tommoor:
Somewhat debugging in public here, it seems like when the selection is collapsed getClientBoundingRect returns zero for all values, if you have an expanded selection it's correct and the bug does not reproduce.

image

src/plugins/sync-plugin.js
dmonad: You are not running standard@v12. And a CI can't fix that. In the future, instead of using your linter, I'm just asking you to run `npx standard --fix` using the shipped version of standard.

I would be very surprised if the linter works differently on your computer. Maybe it does. Maybe you setup some custom rules. Then please make my life easier by fixing the linting manually.

Your code is technically valid `standard` code. So a CI can't enforce you not to run your linter. 

I'm all for setting up CI. For different reasons though.

dmonad:
That seems very odd..

src/plugins/sync-plugin.js
```diff
@@ -228,16 +228,24 @@ export class ProsemirrorBinding {
...
 
     const range = this.prosemirrorView._root.createRange()
     range.setStart(selection.anchorNode, selection.anchorOffset)
     range.setEnd(selection.focusNode, selection.focusOffset)
 
+    // This is a workaround for an edgecase where getBoundingClientRect will
+    // return zero values if the selection is collapsed at the start of a newline
+    // see reference here: https://stackoverflow.com/a/59780954
+    const rects = range.getClientRects()
+    if (!rects.length) {
```
tommoor, line 241: Because this logic is only executed when the selection range is "empty" it seems like a safe addition

src/plugins/sync-plugin.js
tommoor: I don't want to seem like I'm arguing, I just want to be clear – I'm absolutely running `standard@v12`, it's installed in `node_modules` as per the package.json and running `npx standard --fix` from the command line results in this exact same output 🤷

dmonad: It seems to be related to your system. I'm running these packages on Linux, Mac, and Windows with virtually any Node version (10, 12, and 14). Here, I also added travis CI to this project:

[Build Status](https://travis-ci.com/yjs/y-prosemirror)

The project is linted correctly, and I'd like not to change this space everywhere. In that case please don't run the --fix command. Maybe you can figure out the problem and report an issue to the `standard` project.

dmonad:
Thanks for this PR! I copied the relevant changes and published a new release. The issue should be fixed in y-prosemirror@1.0.5

PR #35, opened 2021-01-28
Fixed event listener memory leak

jacobdoran:
Event listeners should be removed when plugin is destroyed.


dmonad:
Right, thank you!

tommoor:
qq @jacobdoran – do you remember if you saw any odd behavior from this bug or was it purely a memory leak fix?

jacobdoran:
> qq @jacobdoran – do you remember if you saw any odd behavior from this bug or was it purely a memory leak fix?

Behavior was fine but it was showing errors in the console log.

PR #37, opened 2021-02-01
Prevent one client syncing private attributes with other clients.

fuzhuyi:
Define yPrivate attribute in prosemirror schema, and prevent one client syncing the private attributes with other clients.


dmonad (changes requested):
Thanks for this @fuzhuyi ,

The PR looks fine, but could you please make sure that `npm test` still works?

src/plugins/sync-plugin.js
```diff
@@ -438,7 +484,25 @@ const createNodeFromYElement = (el, schema, mapping, snapshot, prevSnapshot, com
...
+    if (oldNodeMapping) {
+      const oldNode = oldNodeMapping.get(el)
+      const privateAttrs = findPrivateAttrs(oldNode)
+      for (const i in privateAttrs) {
+        attrs[i] = privateAttrs[i]
+      }
+    }
+
+    let node
+    if (typeof getFallbackNode === 'function') {
```
dmonad, line 497: This can be simplified to a simple if/else construct.

src/plugins/sync-plugin.js
```diff
@@ -711,7 +776,7 @@ export const updateYFragment = (y, yDomFragment, pNode, mapping) => {
     }
     // remove all keys that are no longer in pAttrs
     for (const key in yDomAttrs) {
-      if (pAttrs[key] === undefined) {
+      if (key in pAttrs && pAttrs[key] === undefined) {
```
dmonad, line 779: Please see my comment from your last PR. This condition can never be true.

src/plugins/sync-plugin.js
```diff
@@ -757,7 +822,13 @@ export const updateYFragment = (y, yDomFragment, pNode, mapping) => {
       const leftP = pChildren[left]
       const rightY = yChildren[yChildCnt - right - 1]
       const rightP = pChildren[pChildCnt - right - 1]
-      if (leftY instanceof Y.XmlText && leftP instanceof Array) {
+      if (!(leftP instanceof Array) && isFallbackNode && isFallbackNode(leftP)) {
```
dmonad, line 825: This could be simplified to simply: `if (isFallbackNode(leftP)) {..}`

src/plugins/sync-plugin.js
```diff
@@ -757,7 +822,13 @@ export const updateYFragment = (y, yDomFragment, pNode, mapping) => {
...
+        if (mapping.get(leftY) !== leftP) {
+          yDomFragment.delete(left, 1)
+          yDomFragment.insert(left, /** @type Array{Y.XmlFragment} */ [leftY])
```
dmonad, line 828: I don't really see why it is necessary to delete and insert the same node. Also, it is not possible to insert the same shared type (`leftY`) again in the document. Even if it has been deleted before.

src/plugins/sync-plugin.js
```diff
@@ -424,10 +467,13 @@ const createNodeFromYElement = (el, schema, mapping, snapshot, prevSnapshot, com
       }
     }
   }
-  if (snapshot === undefined || prevSnapshot === undefined) {
-    el.toArray().forEach(createChildren)
-  } else {
-    Y.typeListToArraySnapshot(el, new Y.Snapshot(prevSnapshot.ds, snapshot.sv)).forEach(createChildren)
+  let isKnownNode = !!schema.nodes[el.nodeName]
+  if (isKnownNode) {
```
dmonad, line 471: You could simply do `if (schema.nodes[el.nodeName])`.

But you could simplify this even more by doing the following right at the beginning of `createNodeFromYElement`.

```js
if (!schema.nodes[el.nodeName]) {
  let fallbackNode = schema.node(el.nodeName, attrs, children);
  mapping.set(el, fallbackNode);
  return fallbackNode
}
```

dmonad (changes requested):

PR #41, opened 2021-03-06
Update lib.js

### LINKED ISSUE #40: Support for Vite.js
**Describe the bug**
When using the next generation esm bundler [vite](https://github.com/vitejs/vite), I'm getting this error:

```bash
 > node_modules/y-prosemirror/src/lib.js:1:26: error: No matching export for import "ProsemirrorMapping"
    1 │ import { updateYFragment, ProsemirrorMapping } from './plugins/sync-plugin.js' // eslint-disable-line
      ╵                           ~~~~~~~~~~~~~~~~~~
```

I think this is because `ProsemirrorMapping` is a type generated in the build process. Since `vite` (esbuild) is very strict and uses the source files, it can't find this export. Here is a [comment from evan you](https://github.com/vitejs/vite/issues/1652#issuecomment-768560966) about that.

Do you have any idea how to fix that on your side?

**To Reproduce**
1. import `y-prosemirror` in a vite project
2. start dev server

**Expected behavior**
It should not throw an error

holtwick:
Remove ProsemirrorMapping import. It is unused and causes problems with vite.js. Fixes  #40

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1,4 +1,4 @@
-import { updateYFragment, ProsemirrorMapping } from './plugins/sync-plugin.js' // eslint-disable-line
+import { updateYFragment } from './plugins/sync-plugin.js' // eslint-disable-line
 import * as Y from 'yjs'
 import { EditorView } from 'prosemirror-view' // eslint-disable-line
 import { Node, Schema } from 'prosemirror-model' // eslint-disable-line


dmonad:
Thanks, but simply removing the import will break the build because ProsemirrorMapping is used as a type.

PR #45, opened 2021-04-14
Clean up yDoc observers on PM plugin destroy

AdventureBeard:
I have a few situations where I'd like to keep an already built yDoc in memory (for example, a large one that took a while to parse) and re-use it later. The sync-plugin, however, doesn't seem to expect that behavior, and doesn't clean up the transaction observers on the Prosemirror Plugin destroy hook. This leads to duplicate work and increased processing time on each subsequent re-use.

In this change, I save the transaction hooks as local properties in the ProsemirrorBinding, and remove them in the destroy lifecycle method.


dmonad:
Of course, thank you! I'm going to publish a new release in the next hour.

Issue #34, opened 2020-12-10
Did not deal with the problem of overlapping marks with same name 


dongtc:
**Describe the bug**
Prosemirror supports overlapping marks with same name(as long as they have different attributes):

> excludes: ?⁠string
> Determines which other marks this mark can coexist with. Should be a space-separated strings naming other marks or groups of marks. When a mark is added to a set, all marks that it excludes are removed in the process. If the set contains any mark that excludes the new mark but is not, itself, excluded by the new mark, the mark can not be added an the set. You can use the value "_" to indicate that the mark excludes all marks in the schema.Defaults to only being exclusive with marks of the same type. **You can set it to an empty string (or any string not containing the mark's own name) to allow multiple marks of a given type to coexist (as long as they have different attributes)**.

https://prosemirror.net/docs/ref/#model.MarkSpec.excludes

But prosemirror-binding seems didn't handle it well, i have found the problem code line: 
https://github.com/yjs/y-prosemirror/blob/25cea84874eace50745fd2433847aabceef92b65/src/plugins/sync-plugin.js#L670

If a prosemirror text with the marks:`[{type: 'comment', attrs: {id: 1}}, {type: 'comment', attrs: {id: 2}}]`

It will be convert to:`{comment: {id: 2}}`

You can see lost a mark.

dmonad:
Right.. that makes sense. But this type of information can currently not be properly represented in Yjs / Y.Text.

dongtc:
In my project, i have a temporary solution. The main code is as follows:

![image](https://user-images.githubusercontent.com/13464443/101870330-59b72f80-3bbc-11eb-9169-64a1218e5641.png)

Meantime, i change `createTextNodesFromYText` and `equalYTextPText`

It seems hack but works fine, do you have a better suggest? @dmonad

dmonad:
Hi @dongtc , 

the approach might work, but it will break all existing bindings. I would accept a PR that is well tested and is backwards-compatible. I think it could be possible to use arrays as "attributes" when multiple marks with the same name are present.

dongtc:
Tanks for your suggestion, i will create a PR later. @dmonad

bdbch:
Was this ever fixed? I think I ran into the same issue with overlapping comment marks. Works fine as long as I don't reload my editor but as soon as I reload, the overlapping marks are replaced with the outermost mark taking priority.

dmonad:
For comments, I recommend annotating ranges using relative-positions.

[...]

Issue #51, opened 2021-06-08
Tests occassionally fail


YousefED:
While working on a different issue, I noticed the tests of y-prosemirror occasionally fail, possibly indicating a bug.

To reproduce, simply run the tests multiple times, or change the `testRepeatGenerateProsemirrorChanges300` to run 10 times:

```
export const testRepeatGenerateProsemirrorChanges300 = tc => {
  for (let i = 0; i < 10; i++) {
    checkResult(applyRandomTests(tc, pmChanges, 300, createNewProsemirrorView))
  }
}
```

vivaxy:
Same issue here.

PR #52, opened 2021-06-08
add support for overlapping marks

YousefED:
This addresses issue #34 - while remaining backwards compatible.

There are still two TODO items open open:

- [ ] the test currently fails, I think because there is a yChange attr on one of the documents. I'm not sure what the intention behind yChange is, so I'm not sure what the right way is to fix this
- [ ] the current solution breaks YXmlText.toString because it doesn't handle arrays. I think we should add support for this directly in `yjs`, but want to make sure the solution I'm implementing here is a right approach.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -473,7 +473,15 @@ const createTextNodesFromYText = (text, schema, mapping, snapshot, prevSnapshot,
       const delta = deltas[i]
       const marks = []
       for (const markName in delta.attributes) {
-        marks.push(schema.mark(markName, delta.attributes[markName]))
+        if (Array.isArray(delta.attributes[markName])) {
+          // multiple marks of same type
+          delta.attributes[markName].forEach(attrs => {
+            marks.push(schema.mark(markName, attrs))
+          })
+        } else {
+          // single mark
+          marks.push(schema.mark(markName, delta.attributes[markName]))
+        }
       }
       nodes.push(schema.text(delta.insert, marks))
     }
@@ -545,6 +553,14 @@ const equalAttrs = (pattrs, yattrs) => {
   return eq
 }
 
+const containsEqualMark = (pattrs, yattrs) => {
+  if (Array.isArray(yattrs)) {
+    return !!yattrs.find(el => equalAttrs(pattrs, el))
+  } else {
+    return equalAttrs(pattrs, yattrs)
+  }
+}
+
 /**
  * @typedef {Array<Array<PModel.Node>|PModel.Node>} NormalizedPNodeContent
  */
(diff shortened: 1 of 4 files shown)


src/plugins/sync-plugin.js
```diff
@@ -572,13 +588,25 @@ const normalizePNodeContent = pnode => {
...
+}
+
 /**
  * @param {Y.XmlText} ytext
  * @param {Array<any>} ptexts
  */
 const equalYTextPText = (ytext, ptexts) => {
   const delta = ytext.toDelta()
-  return delta.length === ptexts.length && delta.every((d, i) => d.insert === /** @type {any} */ (ptexts[i]).text && object.keys(d.attributes || {}).length === ptexts[i].marks.length && ptexts[i].marks.every(mark => equalAttrs(d.attributes[mark.type.name] || {}, mark.attrs)))
+  return delta.length === ptexts.length && delta.every((d, i) => d.insert === /** @type {any} */ (ptexts[i]).text && countYTextMarks(d.attributes || {}) === ptexts[i].marks.length && ptexts[i].marks.every(mark => containsEqualMark(d.attributes[mark.type.name] || {}, mark.attrs)))
```
SamDuvall, line 609: Thanks for the PR for handling multiple marks!

I was pulling this PR into my project and ran into an issue with `containsEqualMark`. It looks like you're calling `containsEqualMark` with `yattrs, pattrs` instead of `pattrs, yattrs`.

IdrissMahjoubi:
The first TODO is an open issue I think

>  the test currently fails, I think because there is a yChange attr on one of the documents. I'm not sure what the intention behind yChange is, so I'm not sure what the right way is to fix this

Check: https://github.com/yjs/y-prosemirror/issues/77 and https://github.com/yjs/y-prosemirror/issues/51

vitmf:
Since `ychange`  is lost in the conversion between prosemirror and ydoc, perhaps we should test the marks directly? https://github.com/yjs/y-prosemirror/blob/7a3afbc0d8de35bc01bd63bfc9911318c115c4f7/src/plugins/sync-plugin.js#L961

vadimdemedes:
Hey @dmonad, I know it's an old PR, but could you please let us know if this is the right direction to take for supporting overlapping ProseMirror marks?

piyush-kroolo:
@YousefED @dmonad 

> I know it's an old PR, but could you please let us know if this is the right direction to take for supporting overlapping ProseMirror marks?

+1

dmonad:
Thanks for this PR @YousefED !
I was on the fence about this approach because it does not support concurrently adding overlapping marks (they would overwrite each other).

I implemented a different approach to support overlapping marks in v1.3.0 that supports concurrent formatting. 

https://github.com/yjs/y-prosemirror/compare/fa90ea7f2d787c8c8231be85f02a515bd2d3604c...1b5bd96dd002efa3e75393792e0173209e72e06e

PR #53, opened 2021-06-09
Added method to update a Y.XMLFragment from a Prosemirror Doc Node

SamDuvall:
If the `y-prosemirror` utility functions were called from a `Y.XMLElement` perspective, instead of a `Y.Doc` with a xmlFragment key perspective, developers could re-use the utility functions to perform Y.XMLElement updates on the server. This would give developers another tool to build documents on a server before they are served to clients.

```
export const appendNodes = (
  yXmlFragment: Y.XmlFragment,
  jsonNodes: XmlNode[],
  schema: Schema
) => {
  // Create the initial state
  const json = yXmlFragmentToProsemirrorJSON(yXmlFragment)
  const doc = Node.fromJSON(schema, json)
  const state = EditorState.create({ doc, schema })
  
  // Create the new state
  const nodes = jsonNodes.map(node => Node.fromJSON(schema, node))
  const endPosition = state.doc.content.size
  const tr = state.tr.insert(endPosition, nodes)
  const newState = state.apply(tr)

  // Update the yDoc
  updateYXmlFragmentFromProsemirror(type, newState.doc)
}

```


SamDuvall:
I've had a slight change of heart from my initial comment. I really only needed `updateYFragment` export from the `sync-plugin`.

However, I do think the signature for `yDocToProsemirrorJSON` isn't quite right, because I have a use case where my `Y.XmlFragment` is not hanging off of the `Y.Doc`, which I think is a valid use case. I exported another function `yXmlFragmentToProsemirrorJSON` which converts a `Y.XmlFragment` to JSON and then had `yDocToProsemirrorJSON` use `yXmlFragmentToProsemirrorJSON` to preserve backward-compatibility and not add any duplication.

src/y-prosemirror.js
```diff
@@ -1,5 +1,5 @@
 export * from './plugins/cursor-plugin.js'
-export { ySyncPlugin, isVisible, getRelativeSelection, ProsemirrorBinding } from './plugins/sync-plugin.js'
+export { ySyncPlugin, isVisible, getRelativeSelection, updateYFragment, ProsemirrorBinding } from './plugins/sync-plugin.js'
```
dmonad, line 2: Do you need `updateYFragment`? I don't think this should be exported.

dmonad (changes requested):
Looks good. Please just clarify this one thing.

src/y-prosemirror.js
SamDuvall: No, what I really want is `prosemirrorJSONToYFragment` instead of `prosemirrorJSONToYDoc`. Should I make a separate PR or amend this one?

BrianHung: Just wondering, would `prosemirrorJSONToYFragment` also allow for the conversion of json to `Y.XMLElement`? 

I'm looking through `y-prosemirror` looking for blockers preventing people from using `Y.XMLElement` as the type (in the case of a custom top document), and noticing `prosemirrorJSONToYDoc` hardcodes the `Y.XMLFragment` with `ydoc.get(xmlFragment, Y.XmlFragment)`.

Right now, one can't make a `prosemirrorJSONToYXmlElement` because `updateYFragment` isn't exported as you pointed out.

tommoor: `updateYFragment` exported with a slightly cleaner interface would be incredibly useful. Perhaps something with the signature of: `applyProsemirrorToYDoc(ydoc, prosemirrorDoc)`

PR #54, opened 2021-06-18
Fix `binding === null` due to asynchronous creation. Fixes #42

### LINKED ISSUE #42: TypeError: Cannot read property 'mapping' of null
**Description of the bug**

I am currently using the prosemirror editor together with Yjs in an application made in React. In this application I have developed my own component that creates a tiptap editor (uses prosemirror, I was guided by the examples recommended in the documentation), this since I need to be able to add and remove editors dynamically. The error occurs on line 127 of the file "cursors-plugin.js" of the library "y-prosemirror" located in `"[...]/node_modules/y-prosemirror/src/plugins/cursor-plugin.js: 127"`. And it happens when a user adds a new editor (input) while another user is editing in another editor (input).

**To reproduce**
Due to the context, it is a case that is perhaps difficult to reproduce, however broadly the steps to reproduce the behavior are:
1. Create a project in React.
2. Create a context that stores the y-document and provider.
3. Create a component that instantiates a new TipTap editor in the document.
4. Generate for each element of an array (it can be an array with numbers) a component (the one that was created previously).
5. Create a button to add a new element to the arrangement.
6. Open two tabs in the browser.
7. In a tab, leave the cursor inside one of the editors already created according to the arrangement.
8. In the other tab, press the button that adds a new element to the arrangement.
9. See the error in the console.

** Expected behavior **
The desired behavior is that if a user is editing a field and another user adds a new element to an array (thus a new editor is generated), the application does not break.

**Screenshots**

image

**Environment Information**
 - Browser: Chrome.
 - Back-end: Node.js v12.18.3
 - The dependencies used in this project are:
 - "dependencies": {
    "@material-ui/core": "^4.11.3",
    "@material-ui/icons": "^4.11.2",
    "@material-ui/lab": "^4.0.0-alpha.57",
    "@testing-library/jest-dom": "^5.11.9",
    "@testing-library/react": "^11.2.3",
    "@testing-library/user-event": "^12.6.2",
    "@tiptap/core": "^2.0.0-beta.5",
    "@tiptap/extension-character-count": "^2.0.0-beta.1",
    "@tiptap/extension-collaboration": "^2.0.0-beta.4",
    "@tiptap/extension-collaboration-cursor": "^2.0.0-beta.4",
    "@tiptap/extension-document": "^2.0.0-beta.1",
    "@tiptap/extension-paragraph": "^2.0.0-beta.1",
    "@tiptap/extension-text": "^2.0.0-beta.1",
    "@tiptap/react": "^2.0.0-beta.4",
    "@tiptap/starter-kit": "^2.0.0-beta.4",
    "chart.js": "^2.9.4",
    "query-string": "^6.14.0",
    "react": "^17.0.1",
    "react-dom": "^17.0.1",
    "react-indiana-drag-scroll": "^1.8.0",
    "react-query": "^3.6.0",
    "react-router-dom": "^5.2.0",
    "react-scripts": "4.0.1",
    "socket.io-client": "^3.1.1",
    "web-vitals": "^0.2.4",
    "y-websocket": "^1.3.11",
    "yjs": "^13.5.3"
  },

**Proposed Solution**
The solution that worked for me has been to change line 127 of the file `"[...]/node_modules/y-prosemirror/src/plugins/cursor-plugin.js"` which was like this:

`} else if (current.cursor! = null && relativePositionToAbsolutePosition (ystate.doc, ystate.type, Y.createRelativePositionFromJSON (current.cursor.anchor), ystate.binding.mapping)! == null) {`

Adding the validation "ystate.binding! = Null":

`} else if (current.cursor! = null && ystate.binding! = null && relativePositionToAbsolutePosition (ystate.doc, ystate.type, Y.createRelativePositionFromJSON (current.cursor.anchor), ystate.binding.mapping)! == null) {`

However, I would like to have an answer or opinion as to whether what I just did is okay or not. Thank you very much in advance for your time.

whawker:


src/plugins/sync-plugin.js
```diff
@@ -85,58 +137,21 @@ export const ySyncPlugin = (yXmlFragment, { colors = defaultColors, colorMapping
...
+        return pluginState
       },
       apply: (tr, pluginState) => {
-        const change = tr.getMeta(ySyncPluginKey)
-        if (change !== undefined) {
-          pluginState = Object.assign({}, pluginState)
-          for (const key in change) {
-            pluginState[key] = change[key]
-          }
-        }
```
whawker, line 107: I have removed this, I'm unsure of the larger impact, but everything seems to work fine without it in my project?

src/plugins/sync-plugin.js
```diff
@@ -65,6 +65,57 @@ const getUserColor = (colorMapping, colors, user) => {
...
+    }
+    this.binding = new ProsemirrorBinding(this.type, view)
+    return this
+  }
+
+  destroy () {
+    if (this.binding) {
+      this.binding.destroy()
+    }
+    this.binding = null
```
ankon, line 92: Mostly for consistency: Wouldn't hurt to also explicitly init `this.binding = null` in the constructor?

whawker: Good spot, fixed.

ronnyroeller:
@dmonad We're relying on this fix (currently running on a fork of y-prosemirror). Is there a chance to get this PR merged into `master`?

whawker:
@dmonad Any chance you could take a look over this?

flaviouk:
Can also confirm that this fixes an issue related to the binding null I was having.

src/plugins/sync-plugin.js
dmonad: Other plugins might use this to set properties like `snapshot` or `colorMapping`. We can't just remove this.

dmonad (changes requested):
Thanks @whawker for providing a fix. 

First I want to say that I agree that y-prosemirror really needs a refactor. It was the first ProseMirror plugin I ever wrote (without understanding how ProseMirror works) and it grew significantly over time. A lot of the functionality (e.g. snapshots) are not documented but are widely used by different projects because they found my demos and figured out the  API that I wanted to hide.

This PR will definitely break other people's code. The y-sync plugin state is supposed to be an Object (although I agree that it should have been an instance of YSyncState, but that wasn't clear to me at the time when I wrote the code). Also, you break some of the features (e.g. adding/overwriting properties to the y-sync state). Can you please revert the changes that are not necessary for this PR? It should be easy to review for me.

src/plugins/sync-plugin.js
whawker: @dmonad I have restored this, but dropped the line `pluginState = Object.assign({}, pluginState)`

I guess this is to make the plugin state immutable? In [the example](https://prosemirror.net/examples/track/) I copied this pattern from, it looks like the plugin apply step returns a new instance of the state to achieve this immutability.

However with the asynchronous setting of the `binding` using the view, I'm not 100% sure how to achieve this, anybody have any suggestions?

PR #55, opened 2021-06-21
Add the ability to customise user cursor selection

whawker:
* Exposes a class name for CSS styling
* Add the ability to set a custom selection color or a custom alpha

This PR adds `tinycolor2` as a dependency, meaning
* Any color notation can be used - only 6 character Hex codes worked before
* Alpha transparency can be configured, and will work regardless of the color notation used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3201,6 +3201,11 @@
       "integrity": "sha1-DdTJ/6q8NXlgsbckEV1+Doai4fU=",
       "dev": true
     },
+    "tinycolor2": {
+      "version": "1.4.2",
+      "resolved": "https://registry.npmjs.org/tinycolor2/-/tinycolor2-1.4.2.tgz",
+      "integrity": "sha512-vJhccZPs965sV/L2sU4oRQVAos0pQXwsvTLkWYdqJ+a8Q5kPFzJTuOFwy7UniPli44NKQGAglksjvOcpo95aZA=="
+    },
     "tmp": {
       "version": "0.0.33",
       "resolved": "https://registry.npmjs.org/tmp/-/tmp-0.0.33.tgz",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
   },
   "homepage": "https://github.com/yjs/y-prosemirror#readme",
   "dependencies": {
-    "lib0": "^0.2.34"
+    "lib0": "^0.2.34",
+    "tinycolor2": "^1.4.2"
   },
   "peerDependencies": {
     "yjs": "^13.3.2",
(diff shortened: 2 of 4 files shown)


ankon (approved):
I think it would be good to provide a reason for using another external library for what is probably a fairly simple color operation (`if color has alpha, then use that, otherwise set to provided value`) in the PR description for the future reviewers; other than that -- LGTM.

nc:
+1 - would love to see this merged in :)

ankon:
One problem with the proposed change is that it requires the color of the client cursors to be consistent across the clients. That's not always needed (I don't care whether I appear as "red" or "blue" on your side, for example), and sometimes problematic: The color your client chose might actually be unreadable on my screen because I use a different theme. The latter part can easily be seen when assuming a dark/light theme switch for a collaborative application. 

IMHO at least the actual color selection should be ran through a (optional) local translation layer, which would take the user state and return the actual colors to be used.

tommoor:
Would prefer not to see this additional dependency merged, you can already pass your own builder to the cursor plugin and use any logic you want to store and manipulate the color

ankon:
I think indeed that the cursor builder extensibility is enough for configuring the cursor. This PR is not so much about the _cursor_ though, but rather the hard-coded logic for building up the _selection_.

But, I do agree that simply bringing in more options is not helpful. The plugin can provide the infrastructure, but ideally we want the app/user to use that infrastructure to configure the world. As an alternative, what about closing this PR, and instead creating one that adds a `SelectionBuilder` + `defaultSelectionBuilder` as another (optional) option similar to the cursor builder?

dmonad:
Hi @whawker , 

thank you for this! I'm strict about not adding third-party dependencies to any of my projects (unless they are strictly necessary). Furthermore, I think what you want to achieve can be done without adding a color library to y-prosemirror. @ankon Just added a really nice PR that allows you to manipulate colors or build your own selection based on custom color libraries: https://github.com/yjs/y-prosemirror/pull/94

ronnyroeller:
Thanks for all your guidance, @dmonad !

To close the loop:
We found out that we indeed didn't need the proposed library change. In fact, we even ended up not synchronizing cursor colors at all, avoiding many hard corner cases.
For future readers stumbling over the same issue, we shared the details here: https://medium.com/collaborne-engineering/cursor-colors-with-prosemirror-yjs-76a04b836566

dmonad:
Sweet, thanks for writing this! 

I also found your other article about serverless y-webrtc and added it to the docs. Seems super useful!
https://github.com/yjs/y-webrtc/commit/f100613f417a31aef13f2b1daff0ec544ff0ccff

PR #63, opened 2021-08-16
support marks on element nodes

BrianHung:


BrianHung:
An alternative serialization method might be to just use `setAttribute('marks', node.marks)` to distinguish between `attrs` and `marks. This would also allow `marks` to retain its expected properties of being an array (which help with https://github.com/yjs/y-prosemirror/pull/52), and we wouldn't also have to rely on `schema` to tell if something is a `mark` or `attrs`.

PR #64, opened 2021-08-17
modularize `yDocToProsemirrorJSON` to `Y.XmlText` and `Y.XmlElement`

BrianHung:


src/plugins/sync-plugin.js
```diff
@@ -469,8 +469,7 @@ const createTextNodesFromYText = (text, schema, mapping, snapshot, prevSnapshot,
   const nodes = []
   const deltas = text.toDelta(snapshot, prevSnapshot, computeYChange)
   try {
-    for (let i = 0; i < deltas.length; i++) {
-      const delta = deltas[i]
+    for (const delta in deltas) {
```
dmonad: Please don't use for-in loops for arrays as they fail to transpile in old typescript versions. I deliberately only use normal for loops.

src/lib.js
```diff
@@ -214,95 +213,107 @@ export function prosemirrorToYDoc (doc, xmlFragment = 'prosemirror') {
...
  */
-export function yDocToProsemirrorJSON (
-  ydoc,
-  xmlFragment = 'prosemirror'
-) {
+export function yDocToProsemirrorJSON (ydoc, xmlFragment = 'prosemirror') {
   const items = ydoc.getXmlFragment(xmlFragment).toArray()
+  return {
+    type: 'doc',
+    content: items.map(typeToProseMirrorJSON).flat()
```
dmonad, line 249: The `.flat()` method is a fairly recent addition to EcmaScript. Please use `import { flatten } from 'lib0/array'` instead.

dmonad (changes requested):
Hi @BrianHung,

thanks for working on this. I will give this PR another review if you leave the existing API functional. Unless there is a good reason, we can't change the arguments or the names of the functions without breaking existing code.

SamDuvall:
@BrianHung, @dmonad, I don't think `yDocToProsemirrorJSON` is the correct signature to convert a `Y.XmlFragment` to JSON. In my mind, a `Y.XmlFragment` is an array of `Node`s, not a `Node` with type 'doc'.

See my PR here: https://github.com/yjs/y-prosemirror/pull/53

BrianHung:
> I don't think `yDocToProsemirrorJSON` is the correct signature to convert a `Y.XmlFragment` to JSON. In my mind, a `Y.XmlFragment` is an array of `Node`s, not a `Node` with type 'doc'.

@SamDuvall I agree with both of your points about separating `yDoc` from `Y.XmlFragment`. `YXmlFragmentToProsemirrorJSON` is pretty simple with the newly modularized functions -- just needed that `serialize` function to be outside of it -- so I've added that as well.

dmonad:
Thanks to @BrianHung  and @SamDuvall  we now have two competing implementations :thinking: 

https://github.com/yjs/y-prosemirror/pull/53
https://github.com/yjs/y-prosemirror/pull/64

After reviewing this I think you are right that we should change the signature. I would be fine with making another major release for this feature. Could you both please communicate and come up with a solution that works for both of you. I'll merge the version that you both reviewed and agree with. Thanks!

SamDuvall:
This PR has everything I need for converting `Y.Xml*` to JSON. I am happy to abandon https://github.com/yjs/y-prosemirror/pull/53 and then create a new PR that focuses on JSON to `X.XmlFragment`.

BrianHung:
> we should change the signature

To clarify, this would mean in essence deprecating `yDocToProsemirrorJSON` in favor of the other two utility methods `YXmlFragmentToProsemirrorJSON` and `YXmlElementToProsemirrorJSON`?

Because if we change the signature of `yDocToProsemirrorJSON` to return "an array of Nodes" as Sam suggested, that would be equivalent to `YXmlFragmentToProsemirrorJSON`. 

In my view, deprecating `yDocToProsemirrorJSON` would be good because it'd force the end user to explicitly work with `YXMLFragment` or `YXMLElement` instead of that being abstracted behind `Y.Doc` and a `field`. 

Those using the new API but want the old abstraction of `yDocToProsemirrorJSON` can easily migrate by:
```js
// Previously
let json = yDocToProsemirrorJSON(ydoc)

// Now
let xmlFragment = ydoc.getXmlFragment('prosemirror')
let json = {type: 'doc', content: YXmlFragmentToProsemirrorJSON(xmlFragment)}
```

SamDuvall:
If we're doing a major version bump, then I agree with @BrianHung's proposal to just delete `yDocToProsemirrorJSON`, so long as there is a clear explanation in the release notes.

@dmonad, if we did a major version bump, would you be okay with taking https://github.com/yjs/y-prosemirror/issues/34 as-is or would you still like a PR that is backwards-compatible? I'm coming across some issues with annotations/comment/suggestions that would benefit greatly from marks of the same type (e.g. comment) being allowed.

flaviouk (approved):
This would be a great addition!

I also need to be able to set prosemirror content json to a YFragment, I believe updateYFragment does this, but I'm curious if we should create a separate PR for that:
https://github.com/yjs/y-prosemirror/pull/53

EricHasegawa:
Any updates here? Having the same issue needing overlapping nodes with different attributes

PR #67, opened 2021-09-07
Change Y.XmlElement instanceof check to constructor.name

TeemuKoivisto:
In local development, at times you get multiple instances of the same library by npm linking modules
back and forth. Using instanceof with two Yjs libraries loaded fails which would seem unnecessary or at least
should be handled somewhere else.

Related https://discuss.yjs.dev/t/prosemirror-build-variations-and-yjs-shared-cursor/228/3

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -390,7 +390,7 @@ export class ProsemirrorBinding {
 const createNodeIfNotExists = (el, schema, mapping, snapshot, prevSnapshot, computeYChange) => {
   const node = /** @type {PModel.Node} */ (mapping.get(el))
   if (node === undefined) {
-    if (el instanceof Y.XmlElement) {
+    if (el.constructor.name === Y.XmlElement.name) {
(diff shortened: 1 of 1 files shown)


dmonad:
Comparing the constructors is fairly ingrained into Yjs. This is by far the most performant way to compare types and is therefore preferred over `constructor.name` which basically does a string comparison. This would seriously downgrade performance when we have documents with over a million `Structs` (I need to distinguish between different kinds of content very often).

There might be a more performant way to implement what you are asking for. However, I'm fairly sure that any approach will hit performance significantly.

I don't want to spend this amount of time when I actually think that the fault is with the user of this library. There is no good reason to have two versions of Yjs installed. I understand that tooling often makes it difficult to prevent having the same package installed multiple times. But then again.. maybe the tooling is wrong for enforcing such behavior. This is part of the reason why web applications nowadays are so bloated. 

If you submit a PR that solves this issue and doesn't reduce performance in [crdt-benchmarks](https://github.com/dmonad/crdt-benchmarks) then I'd be happy to merge it.

Issue #56, opened 2021-06-28
Get mapping without sync-plugin


DanSnow:
**Checklist**

[ ] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
[ ] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Is your feature request related to a problem? Please describe.**
I want to use `relativePositionToAbsoultePosition` and `absolutePositionToRelativePosition` at the websocket server side. These helper functions require a mapping parameter. Currently, I just copy and paste [`createNodeFromYElement`](https://github.com/yjs/y-prosemirror/blob/1c393fb3254cc1ed4933e8326b57c1316793122a/src/plugins/sync-plugin.js#L412) and its related functions from sync-plugin.js but since it's private API. I want to ask is it possible to expose a public API to do get the mapping from Y.js document?

**Describe the solution you'd like**
Expose a helper function to get the mapping

**Describe alternatives you've considered**

**Additional context**

jamesopti:
This would be useful for us as well to work with relative positions on the server side.

jjgmckenzie:
Any plans with this in 2023?

PR #69, opened 2021-10-05
Expose the `mapping` when converting Prosemirror (JSON) to Y.Doc instances

ankon:
This mapping is needed for invoking the position transformation functions, and therefore
shouldn't be thrown away.

---

I added the parameter at the end and made it optional with a default matching the previous behavior, so I claim that this should be quite safe.

Possibly related: 
* https://github.com/yjs/y-prosemirror/issues/56
* https://github.com/yjs/y-prosemirror/pull/64 seems to be focused on the other direction to be more reusable, and may trigger a major version update -- so changing the order of the parameters here could also be done as part of that.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -193,16 +193,17 @@ export const relativePositionToAbsolutePosition = (y, documentType, relPos, mapp
  *
  * @param {Node} doc
  * @param {string} xmlFragment
+ * @param {ProsemirrorMapping} mapping
  * @return {Y.Doc}
  */
-export function prosemirrorToYDoc (doc, xmlFragment = 'prosemirror') {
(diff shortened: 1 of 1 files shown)


jamesopti:
We have the need for this too! Thanks for this PR @ankon !

@dmonad any thoughts here?

HelloAlexPan:
Same here! Bump

jamesopti:
Any update on this one?

EricHasegawa:
Bumping this! @dmonad Any thoughts?

ankon:
Purely for the record: I no longer work with prosemirror at all, so if you want to merge this -- please do! But if it needs fixes/adjustments, please also feel absolutely free to do them as I won't be able to.

PR #71, opened 2021-10-16
fix: abort scheduled processes on destruction

### LINKED ISSUE #70: Uncaught TypeError: Cannot read properties of null (reading 'matchesNode')
> Please save me some time and use the following template. In 90% of all issues I can't reproduce the problem because I don't know what exactly you are doing, in which environment, or which y-* version is responsible. Just use the following template even if you think the problem is obvious.

**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
The syncPlugin will run into an error if the editor get's destroyed. It looks like a call to `tiptap#unregisterPlugin` causes every other plugin view to update.
`sync-plugin.js` will then schedule a new `_forceRenderer()` cycle:

```js
// sync-plugin.js

setTimeout(() => {
  binding._forceRerender()
  view.dispatch(view.state.tr.setMeta(ySyncPluginKey, { binding }))
}, 0)
```

When the editor gets destroyed in the meantime, the scheduled function will cause the following error:

```
index.es.js:4914 Uncaught TypeError: Cannot read properties of null (reading 'matchesNode')
    at EditorView.updateStateInner (index.es.js:4914)
    at EditorView.updateState (index.es.js:4885)
    at Editor.dispatchTransaction (tiptap-core.esm.js:3102)
    at EditorView.dispatch (index.es.js:5162)
    at sync-plugin.js:282
    at ProsemirrorBinding.mux (mutex.js:35)
    at ProsemirrorBinding._forceRerender (sync-plugin.js:278)
    at sync-plugin.js:134
```

just noticed,... the same is true for `setMeta` which calls `eventloop.timeout`:

```
index.es.js:4914 Uncaught TypeError: Cannot read properties of null (reading 'matchesNode')
    at EditorView.updateStateInner (index.es.js:4914)
    at EditorView.updateState (index.es.js:4885)
    at Editor.dispatchTransaction (tiptap-core.esm.js:3102)
    at EditorView.dispatch (index.es.js:5162)
    at lib.js:30
    at Map.forEach (<anonymous>)
    at updateMetas (lib.js:25)
```

**To Reproduce**
* unregister a plugin right before the editor get's destroyed
* watch `binding._forceRenderer()` throw an error

**Expected behavior**
`setTimeout` handler should get cleared on destroy.

**Environment Information**
 - Browser: Chromium
 - `"y-prosemirror": "^1.0.9"`
 - `"y-protocols": "^1.0.5"`
 - `"@tiptap/core": "^2.0.0-beta.122"`
 - `"@tiptap/extension-collaboration": "^2.0.0-beta.25"`

**Additional Information**

I currently monkeypatch the plugin:

```ts
const fragment = this.options.fragment
                         ? this.options.fragment
                         : this.options.document.getXmlFragment(this.options.field);
const syncPlugin: Plugin = ySyncPlugin(fragment);
let changedInitialContent = false;
let handler: number;
syncPlugin.spec.view = view => {
    const binding = new ProsemirrorBinding(fragment, view);
    // Make sure this is called in a separate context
    if (handler) {
        clearTimeout(handler);
    }
    handler = setTimeout(() => {
        binding._forceRerender();
        view.dispatch(view.state.tr.setMeta(ySyncPluginKey, { binding }));
    }, 0) as unknown as number;
    return {
        update: () => {
            const pluginState = syncPlugin.getState(view.state);
            if (pluginState.snapshot == null && pluginState.prevSnapshot == null) {
                if (changedInitialContent || view.state.doc.content.findDiffStart(view.state.doc.type.createAndFill().content) !== null) {
                    changedInitialContent = true;
                    binding._prosemirrorChanged(view.state.doc);
                }
            }
        },
        destroy: () => {
            clearTimeout(handler);
            binding.destroy();
        }
    };
};
```

If I got time, I will prepare a PR for it right away :)

hoebbelsB:
fixes #70

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -31,12 +31,17 @@ const updateMetas = () => {
   })
 }
 
+/**
+ * @return TimeoutObject | null
+ */
 export const setMeta = (view, key, value) => {
+  let timeout = null;
   if (!viewsToUpdate) {
     viewsToUpdate = new Map()
-    eventloop.timeout(0, updateMetas)
+    timeout = eventloop.timeout(0, updateMetas)
   }
   map.setIfUndefined(viewsToUpdate, view, map.create).set(key, value)
+  return timeout;
 }
 
 /**
--- a/src/lib.js
+++ b/src/lib.js
@@ -35,13 +35,13 @@ const updateMetas = () => {
  * @return TimeoutObject | null
  */
 export const setMeta = (view, key, value) => {
-  let timeout = null;
+  let timeout = null
   if (!viewsToUpdate) {
     viewsToUpdate = new Map()
     timeout = eventloop.timeout(0, updateMetas)
   }
   map.setIfUndefined(viewsToUpdate, view, map.create).set(key, value)
-  return timeout;
+  return timeout
 }
 
 /**
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -78,68 +78,72 @@ export const createDecorations = (state, awareness, createCursor) => {
  * @param {string} [opts.cursorStateField] By default all editor bindings use the awareness 'cursor' field to propagate cursor information.
  * @return {any}
  */
-export const yCursorPlugin = (awareness, { cursorBuilder = defaultCursorBuilder, getSelection = state => state.selection } = {}, cursorStateField = 'cursor') => new Plugin({
-  key: yCursorPluginKey,
-  state: {
-    init (_, state) {
-      return createDecorations(state, awareness, cursorBuilder)
-    },
-    apply (tr, prevState, oldState, newState) {
-      const ystate = ySyncPluginKey.getState(newState)
-      const yCursorState = tr.getMeta(yCursorPluginKey)
-      if ((ystate && ystate.isChangeOrigin) || (yCursorState && yCursorState.awarenessUpdated)) {
-        return createDecorations(newState, awareness, cursorBuilder)
-      }
-      return prevState.map(tr.mapping, tr.doc)
-    }
-  },
-  props: {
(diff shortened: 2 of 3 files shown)


ValentaTomas:
This fix (y-cursor-plugin part specifically) solved my problem with #70 too - it is an almost constant problem when you allow user to switch between several ProseMirror editors.

Are there some other problems with this PR request that prevent it from being merged?

ValentaTomas:
Shouldn't we also clear `setTimeout` from https://github.com/yjs/y-prosemirror/blob/f593d68a90ec996169718ac33b4c9d35f977ee42/src/plugins/sync-plugin.js#L114 ?

Also how about using `timeout` from `lib0/eventloop`-- the `timeout` is already used in this package when updating cursor.

@dmonad Can you please check this issue? I'm using these fixed in our project, but I would be really glad if I could use the `y-prosemirror` official package again.

tommoor:
This seems like a clean bugfix~ not sure why it's sat for months.

dmonad:
Thank you for the PR @hoebbelsB and sorry that it took so long.

I went for a different approach that checks whether the binding is destroyed. I prefer not to clear timeouts because that leads to further issues in the future. `setMeta`, for example is used in third-party packages as well that don't know that they should clear the timeout.

@ValentaTomas I implemented your suggestion as well and switched from `setTimeout` to `eventloop.timeout`. Thanks for the suggestion

ValentaTomas:
Thank you for your work!

PR #73, opened 2021-10-22
Linking to documentation for Awareness protocol

mitar:
For those like me who are still learning things.


dmonad:
Thank you, improving the docs is much appreciated!

Issue #66, opened 2021-09-06
Add meta to all ysync triggered transactions


TeemuKoivisto:
**Is your feature request related to a problem? Please describe.**
If you want to track transactions and depend on knowing what to intercept and what to pass, having some way of knowing what plugin created what transaction is quite crucial.

**Describe the solution you'd like**
So as of now, the ySync triggers some transactions with its own pluginKey to notify itself to update its state. However, it also triggers other transactions that modify the document without any meta fields added. For example, the `_forceRerender` replaces the whole document and it's quite impossible to detect where it originated from. In my own repo, I added a field `tr.setMeta('_forceRerender', true)` to be able to omit those transactions as non-user originated.

**Describe alternatives you've considered**
The other solution is to add meta to all user-created transactions yourself but that seems quite burdensome and flaky.

**Additional context**
na

dmonad:
The ySync plugin state has a meta property `isChangeOrigin` which is true only if the state was changed by a remote user. Is this what you are looking for?

```js
const { isChangeorigin } = tr.getMeta(ySyncPluginKey)
```

TeemuKoivisto:
No, I dont think so. As I mount the editor and the ySync plugin replaces the PM doc with the ydoc that transaction is triggered without any meta fields. This is an issue as I want to track and in times prevent changes made by the user to some nodes.

TeemuKoivisto:
Update: I have the same problem with the snapshots; without distinguishing meta fields I can't skip them. I guess the most prudent option would be to use the ySyncPluginKey for all transactions originating from the ySync plugin (and do the same for other y-prosemirror plugins if necessary) but instead of passing directy the plugin change in the payload, they would use actions/named events to know what transactions update the plugin state. Or they add some additional "plugin-update" meta-field for the payload. Or just use "true" for transactions that don't update and the regular change objects for those that change.

TeemuKoivisto:
@dmonad if you have time to look at it, I created a PR regarding this. Even after a couple of months I still think this is a valid feature as the other options are less clean on my side eg adding a meta tag to all transactions PM creates and then checking if it doesn't exist instead of just ySyncPluginKey.

TeemuKoivisto:
This was merged by https://github.com/yjs/y-prosemirror/pull/74 and can confirm it works with the latest (v1.0.11) version. Thanks a ton @dmonad!

PR #74, opened 2021-11-05
Use setMeta for all trs dispatched by ySyncPlugin

TeemuKoivisto:
So this is continuation for #66 and it is the fix I have been using with a locally linked version of `y-prosemirror`. Why I need this is because I'm intercepting transactions as in track changes functionality and I can't really detect whether the transaction was originally from ProseMirror or from ySyncPlugin.

With the added booleans in the meta field I can omit those transaction that are caused by for example rendering a snapshot. Also I think it is better DX if you can immediately see from a transaction that "ah this was dispatched by ySyncPlugin". It doesn't seem there are major downsides, without having the meta payload typed it is I guess a little harder to see what are the passed values.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -279,7 +279,7 @@ export class ProsemirrorBinding {
       const fragmentContent = this.type.toArray().map(t => createNodeFromYElement(/** @type {Y.XmlElement} */ (t), this.prosemirrorView.state.schema, this.mapping)).filter(n => n !== null)
       // @ts-ignore
       const tr = this.prosemirrorView.state.tr.replace(0, this.prosemirrorView.state.doc.content.size, new PModel.Slice(new PModel.Fragment(fragmentContent), 0, 0))
-      this.prosemirrorView.dispatch(tr)
+      this.prosemirrorView.dispatch(tr.setMeta(ySyncPluginKey, { isChangeOrigin: true }))
     })
   }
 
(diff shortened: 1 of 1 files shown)


src/plugins/sync-plugin.js
```diff
@@ -279,7 +279,7 @@ export class ProsemirrorBinding {
       const fragmentContent = this.type.toArray().map(t => createNodeFromYElement(/** @type {Y.XmlElement} */ (t), this.prosemirrorView.state.schema, this.mapping)).filter(n => n !== null)
       // @ts-ignore
       const tr = this.prosemirrorView.state.tr.replace(0, this.prosemirrorView.state.doc.content.size, new PModel.Slice(new PModel.Fragment(fragmentContent), 0, 0))
-      this.prosemirrorView.dispatch(tr)
+      this.prosemirrorView.dispatch(tr.setMeta(ySyncPluginKey, true))
```
dmonad, line 282: `true` is not a supported value for `ySyncPluginKey`. However, you can set `tr.setMeta(ySyncPluginKey, { isChangeOrigin })`.

src/plugins/sync-plugin.js
```diff
@@ -99,7 +99,7 @@ export const ySyncPlugin = (yXmlFragment, { colors = defaultColors, colorMapping
       },
       apply: (tr, pluginState) => {
         const change = tr.getMeta(ySyncPluginKey)
-        if (change !== undefined) {
+        if (change !== undefined && typeof change === 'object') {
```
dmonad, line 102: Please remove this change. I don't think it makes sense to introduce the value `true` when there already is a `isChangeOrigin` property.

dmonad (changes requested):
Happy to merge this after the feedback is addressed.

src/plugins/sync-plugin.js
TeemuKoivisto: I updated the payloads to `{ isChangeOrigin: true }`. I assume this is what you intended?

dmonad:
Great, thanks! I published a new release with your patch.

PR #78, opened 2021-11-08
Emit a warning if color is not 6 digit hex

Flamenco:


Flamenco:
To prevent multiple regex calls and logs, I limited the warnings output to a single one.

dmonad:
Great, thanks!

Issue #83, opened 2021-11-17
'Cannot convert undefined or null to object' in equalAttrs


TeemuKoivisto:
In equalAttrs, `yattrs` can be null thus crashing when `Object.keys(yattrs)` is executed.

I suppose this happens because iterating over attribute keys a check is done:

```js
  eq = key === 'ychange' || l === r || (typeof l === 'object' && typeof r === 'object' && equalAttrs(l, r))
```

Which in case when comparing a regular object `{}` against `null` tries to recurse the null because null's typeof is object. A rather unfortunate fact but that's how it is. So maybe a fix could be adding a nullness check in that condition or at the start of the function.

**To Reproduce**
I'll write a reproduction if needed.

**Environment Information**
"y-prosemirror": "^1.0.12",

dmonad:
Sure, you can create a PR for it.

PR #84, opened 2021-11-18
Check for object nullness in equalAttrs

TeemuKoivisto:
As described in https://github.com/yjs/y-prosemirror/issues/83 this is the proposed fix. Not sure would you @dmonad prefer inlining the condition. I used a function since it felt a little cleaner.


dmonad:
Thank you @TeemuKoivisto ! I will publish a new release shortly.

Close #83

PR #87, opened 2021-12-16
Allow passing a custom undoManager to undoPlugin

nc:
So you can have one UndoManager for multiple ProseMirror instances


dmonad:
Makes sense. We don't need this file though. But I'll remove it and make a new release.

ankon:
Just leaving a short comment here for future readers: We also needed this change in our remirror-based collaborative editor, as we observed leaks of the `Y.UndoManager` instances whenever a new state for the editor was created. We move the initialization of the undo manager into the "plugin" to ensure that there is a 1:1 mapping between the `Y.Doc` instance and the `Y.UndoManager`.

I'm not sure whether prosemirror has protections against that (only one call to `init` ever?), or whether the same issue might exist there as well.

PR #91, opened 2022-02-06
Mark all transactions originating from the sync-plugin as "not for history"

ankon:
Changes coming through Yjs are not changes from the local user, but rather they represent
changes in other places in the system. Such changes do not need to be tracked and made
undo-able. This plugin also triggers prosemirror changes when rendering snapshots
and when forcing re-renders, which do a full document replace step.

By marking these with the `"addToHistory"` meta the [prosemirror-history plugin](https://prosemirror.net/docs/ref/#history) will be
able to ignore these changes for the purposes of providing undo/redo functionality on the
editor model level, without affecting a Yjs document level undo/redo.

---
See also https://discuss.yjs.dev/t/avoiding-y-prosemirror-undo/985

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -217,6 +217,15 @@ export class ProsemirrorBinding {
     this._domSelectionInView = null
   }
 
+  /**
+   * Create a transaction for changing the prosemirror state.
+   *
+   * @returns
+   */
+  get _tr () {
+    return this.prosemirrorView.state.tr.setMeta('addToHistory', false)
+  }
+
   _isLocalCursorInView () {
     if (!this.prosemirrorView.hasFocus()) return false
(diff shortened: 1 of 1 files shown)


dmonad:
I'm not sure if this will yield the intended result. It would probably be best to use the Yjs undo-manager and remove the history plugin.

However, I'll add this addition because it's a definite improvement.

ankon:
Thanks a lot!

For us this actually works quite well, and the yjs undo manager was problematic for two reasons:
1. It undo-s on the level of Yjs, meaning user A can undo a change they received from user B. This is at least somewhat surprising to our users.
2. More importantly for us: Yjs undo replaces the prosemirror document completely, which kills our annotations -- these are based on document positions, and when the document content gets replaced the annotations vanish with the replaced document.

Having the option here definitely helps!

PR #92, opened 2022-02-15
Clear timeout when destroying plugin

fson:
Not clearing the timeout causes issues where code is run after the `EditorView` and its plugins have already been destroyed. In our app this resulted in two errors that manifested in different cases where the editor view was being unmounted:
- `Uncaught TypeError: Cannot read properties of null (reading 'matchesNode')`
- `Uncaught RangeError: Applying a mismatched transaction`

Both errors were fixed by applying the monkey patch kindly provided by @hoebbelsB in #70.

This PR implements that fix in `sync-plugin` so that we could stop monkey patching it.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -75,6 +75,7 @@ const getUserColor = (colorMapping, colors, user) => {
  */
 export const ySyncPlugin = (yXmlFragment, { colors = defaultColors, colorMapping = new Map(), permanentUserData = null } = {}) => {
   let changedInitialContent = false
+  let rerenderTimeoutId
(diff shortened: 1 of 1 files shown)


hoebbelsB:
@fson i've already [provided a PR](https://github.com/yjs/y-prosemirror/pull/71) for it, which is open since Oct. 2021 .

just noticed there are merge conflicts, gonna resolve them tonight

dmonad:
Thank you!

I'll merge this because it contains little changes that I can understand easily. I think #70 might fix another issue as well but I currently can't merge it.

I'll release a new version in the coming days with a few bugfixes. I will port the relevant changes from #70 manually.

PR #94, opened 2022-03-01
Extract determining the attributes used for the selection into a configurable "selection builder"

ankon:
This is modelled similar to the existing "cursor builder"; note that with this change the selection
is now tagged with the `ProseMirror-yjs-selection` class as well.

See https://github.com/yjs/y-prosemirror/pull/55#issuecomment-1055169237; I'll would propose to close that PR in favor of this one.

### DIFF
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -25,6 +25,19 @@ export const defaultCursorBuilder = user => {
   return cursor
 }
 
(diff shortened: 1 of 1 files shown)


ankon:
https://github.com/yjs/y-prosemirror/pull/94/checks?check_run_id=5373099760 doesn't seem to be related to my changes?

PR #96, opened 2022-03-17
Support passing clientID to doc transformation utils

gauravtiwari:
Thanks for this library 🙏 

I have been trying to figure out the initial hydration process especially when the editor content is not saved in `Y.Doc()` format

This PR addresses a missing feature to fully support the template engine approach described here: https://discuss.yjs.dev/t/initial-offline-value-of-a-shared-document/465/13

### After this PR

```javascript
// Server
clientID = userSessionID // (only happens once across a collaboration session)
const contentParser = markdownParser(pmSchema)
const pmDoc = contentParser.parse(markdown)
const ydocTemplate = toBase64(encodeStateAsUpdateV2(prosemirrorToYDoc(pmDoc, clientID)))
```

```javascript
// client
const ydoc = new Y.Doc()
// Loaded server side and passed to DOM
Y.applyUpdateV2(ydoc, fromBase64(ydocTemplate))
ydoc.clientID = userSessionID
```

### Before this PR

The `prosemirrorToYDoc` function uses new `Y.Doc()`, which assigns a random clientID causing a mismatch: 

```javascript
const ydoc = new Y.Doc() // assigns a random clientID in the browser 
Y.applyUpdateV2(ydoc, fromBase64(ydocTemplate)) // random clientID on the server

// => store.clients does not include ydoc.clientID
```

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -195,11 +195,17 @@ export const relativePositionToAbsolutePosition = (y, documentType, relPos, mapp
  * collaboration has begun as all history will be lost
  *
  * @param {Node} doc
+ * @param {number} clientID
  * @param {string} xmlFragment
  * @return {Y.Doc}
  */
-export function prosemirrorToYDoc (doc, xmlFragment = 'prosemirror') {
+export function prosemirrorToYDoc (doc, clientID = null, xmlFragment = 'prosemirror') {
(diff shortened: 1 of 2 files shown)


gauravtiwari:
Turns out this isn't an issue anymore and random ids are much better to avoid conflicts

Issue #95, opened 2022-03-06
TypeScript type error with latest yjs


Saul-Mirone:
> Please save me some time and use the following template. In 90% of all issues I can't reproduce the problem because I don't know what exactly you are doing, in which environment, or which y-* version is responsible. Just use the following template even if you think the problem is obvious.

**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
```bash
> tsc --noEmit

../../node_modules/.pnpm/y-prosemirror@1.0.15_y-protocols@1.0.5+yjs@13.5.28/node_modules/y-prosemirror/dist/src/plugins/sync-plugin.d.ts:55:32 - error TS2314: Generic type 'YEvent<T>' requires 1 type argument(s).

55     _typeChanged(events: Array<Y.YEvent>, transaction: Y.Transaction): void;
                                  ~~~~~~~~


Found 1 error in ../../node_modules/.pnpm/y-prosemirror@1.0.15_y-protocols@1.0.5+yjs@13.5.28/node_modules/y-prosemirror/dist/src/plugins/sync-plugin.d.ts:55
```

**To Reproduce**
1. Go to https://github.com/Saul-Mirone/milkdown
2. Check out the `renovate/lock-file-maintenance` branch
3. Run `pnpm install && pnpm build:packs`

**Expected behavior**
Typesciript can compile with no error.

**Screenshots**

**Environment Information**
 - Browser / Node.js [e.g. Chrome, Firefox, Node.js]
 - Yjs@13.5.28, y-prosemirror@1.0.15

**Additional context**

dmonad:
This is likely a typescript issue. Please specify which typescript version are you using.

Saul-Mirone:
A gentle ping @dmonad

PR #99, opened 2022-03-29
fix(type): type parameter errors of yjs

Saul-Mirone:
This PR fix the type error when using:
1. latest y-prosemirror
2. latest yjs
3. has `strcit: true` options enabled for typescript

Closes: #95


dmonad:
Thank you! I will get this released this week with some other updates.

PR #107, opened 2022-05-07
feat: handle composition

vincentdchan:
# Why?

When the user inputs with IME, the middle process will be committed, that's unnecessary. This PR wants to optimize this behaviour. When the editor is in the composing state, Yjs don't need to commit the changes. When the composition was finished, the plugin commits the changes.


dmonad:
Hi @vincentdchan, thank you for this!

I'm wondering how this workaround behaves in a collaborative environment. I expect that remote changes will delete the composing state which is probably not what you want.

Would it be possible for you to test if this works reliably when other people are editing the same document / paragraph?

cqh963852:
I have create a project with y-prosemirror. It use vincentdchan:feat/ime branch.

It has some problem. While delete a some content. The user is under composing, cursor comes wrong.

I have check the shimo's behavior. During composing, It much like a offline editor.

cqh963852:
1. type **123456**
2. image
3. insert between **34** with composing **aa**
4. image

cursor data received by another user
3baeb6547606f644f94dc097a46d0f4

cqh963852:
😀I have implement a custom provider based on socket.io.
While composing, I make sync suspense. Just like disconnect.
After composing, I send a sync message. It works fine.

😔I tried modify the demo in y-prosemirror, but the disconnect action in y-webrtc provider not works fine.

PR #109, opened 2022-05-12
Add on initial apply callback

flaviouk:
Happy to rename if you feel like this should be a different name.

The goal with this is to know when the initial yjs-prosemirror sync has finished so that I can do something else


dmonad:
Thanks, I'll rename this probably to `onFirstRender` (just a personal preference; feels clearer).

Issue #110, opened 2022-05-12
Prevent certain changes from being undone


thatsjonsense:
**Is your feature request related to a problem? Please describe.**
We have some ProseMirror commands that modify the doc programmatically, and we don't want users to be able to undo these. Specifically, when users upload an image, we first put a tempUrl into the state, and then put an uploadedUrl once the upload is incomplete. Right now, a user can hit undo and they'll go back to the tempUrl, which should never be a state you can get stuck in.

**Describe the solution you'd like**
[prosemirror-history has an `addToHistory` meta key](https://github.com/ProseMirror/prosemirror-history/blob/master/src/history.js#L380) you can make on a transaction that prevents undo on a specific transaction. I wish y-prosemirror would also honor the same meta key.

**Describe alternatives you've considered**
The only alternative I can think of is to register listeners on the undo manager, detect when undos happen that change this state, and then reapply them. This feels quite hacky.

**Additional context**

dmonad:
A flag like this would be nice.

I'm currently a bit overloaded with work, so here is my proposal: You create a PR with really nice documentation on how this feature works. Maybe create a Undo section in the README. I'll make some adjustments and implement the API that you suggested.

thatsjonsense:
Hi Kevin that's very fair. I've added a proposed PR for this here: https://github.com/yjs/y-prosemirror/pull/111 , happy to explain more or make any changes if helpful

dmonad:
Implemented in `y-prosemirror@1.1.0`.

PR #111, opened 2022-05-13
Readme section for undo/redo, proposed addToHistory command

thatsjonsense:
- Adds a readme section on undos
- Proposes a way to prevent some changes from being undone per discussion in https://github.com/yjs/y-prosemirror/issues/110

The proposed API mirrors prosemirror-history, for more info see https://github.com/ProseMirror/prosemirror-history/blob/master/src/history.js#L380

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,25 @@ ydoc.applyUpdate(update)
 const node = yDocToProsemirrorJSON(ydoc)
 ```
 
+### Undo/Redo
+
+The package exports `undo` and `redo` commands which can be used in place of
+[prosemirror-history](https://prosemirror.net/docs/ref/#history) by mapping the
+mod-Z/Y keys - see [ProseMirror](https://github.com/yjs/yjs-demos/blob/main/prosemirror/prosemirror.js#L29)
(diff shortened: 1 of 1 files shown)


dmonad:
Thanks :+1: 
I implemented the API and will release it in a bit.

thatsjonsense:
Hey @dmonad , thanks so much for pushing this change. I've been testing it locally though and I'm still not seeing the intended behavior. It looks like when ProseMirror sends multiple transactions in a row, y-prosemirror is batching them together and only checking this meta key for the last one. This means that the whole batch is still actually getting captured.

image

Is there any way to prevent this?

dmonad:
I think this is fixed in `v1.1.1`, can you please verify?

thatsjonsense:
Hey @dmonad unfortunately not. I think the issue remains that when multiple prosemirror transactions happen in short sequence, and Yjs batches them together into its own transaction, only the last prosemirror transaction's addToHistory value is used:

image

thatsjonsense:
I haven't tested this code, but I suspect a solution would look something like:

image

dmonad:
@thatsjonsense I think the next release should fix this issue.

dmonad:
If not, please write a test case so I can reproduce it better.

PR #112, opened 2022-05-16
Fix wrong timeout object

KentoMoriwaki:
I encountered an error `TypeError: Cannot read property 'matchesNode' of null` when using with react and HMR.
(same as https://github.com/ueberdosis/tiptap/issues/1451)

I just found that y-prosemirror dispatches a transaction even after destroyed, and it's due to failed to clear timeout by wrong use of `lib0/eventloop`.


dmonad:
Right, thank you!

Will publish a new release in a bit.

PR #118, opened 2022-05-29
fix (sync plugin)[Issue #43]: Restore relativeSelection using a NodeSelection if appropriate

### LINKED ISSUE #43: Prosemirror SelectedNode Issue
**I checked everywhere and there is no solution to it.**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [ ] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
Prosemirror provides the ability to create NodeSelection. But when content is updated through YJS. The node selection gets removed and become the TextSelection.

**To Reproduce**
1. Go to Y-Prosemirror Demo
2. Insert an Image and click on the Image, the Image shows a border, indication of the NodeSelection.
3. Now if The other collaborator does any change in the document. The border goes away and it creates a TextSelection of that image.
4. This behaviour is very weird because I am showing a toolbar on Image Selection and that gets disappear every time.

**Expected behavior**
Ideally it should retain the NodeSelection and not convert it to TextSelection.

**Environment Information**
Everywhere

**Additional context**
I can detect the changes and move TextSelection back to NodeSelection, that requires an additional update to the state and behaviour will not be smooth at all. That doesn't seem the right way to me. Ideally y-prosemirror should handle this scenario itself.

jamesopti:
## Changes

- Update `getRelativeSelection` to capture whether or not the current selection is a NodeSelection
- Update `restoreRelativeSelection` to use a node selection if restoring one

## Context

Attempts to fix #43

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -4,7 +4,7 @@
 
 import { createMutex } from 'lib0/mutex'
 import * as PModel from 'prosemirror-model'
-import { Plugin, TextSelection } from 'prosemirror-state' // eslint-disable-line
+import { Plugin, NodeSelection, TextSelection } from 'prosemirror-state' // eslint-disable-line
 import * as math from 'lib0/math'
 import * as object from 'lib0/object'
 import * as set from 'lib0/set'
@@ -194,14 +194,18 @@ const restoreRelativeSelection = (tr, relSel, binding) => {
     const anchor = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.anchor, binding.mapping)
     const head = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.head, binding.mapping)
     if (anchor !== null && head !== null) {
-      tr = tr.setSelection(TextSelection.create(tr.doc, anchor, head))
+      const selection = relSel.isNodeSelection
+        ? NodeSelection.create(tr.doc, Math.min(anchor, head))
+        : TextSelection.create(tr.doc, anchor, head)
+      tr = tr.setSelection(selection)
     }
   }
 }
 
 export const getRelativeSelection = (pmbinding, state) => ({
   anchor: absolutePositionToRelativePosition(state.selection.anchor, pmbinding.type, pmbinding.mapping),
-  head: absolutePositionToRelativePosition(state.selection.head, pmbinding.type, pmbinding.mapping)
+  head: absolutePositionToRelativePosition(state.selection.head, pmbinding.type, pmbinding.mapping),
+  isNodeSelection: state.selection instanceof NodeSelection,
 })
 
 /**
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -194,7 +194,7 @@ const restoreRelativeSelection = (tr, relSel, binding) => {
     const anchor = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.anchor, binding.mapping)
     const head = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.head, binding.mapping)
     if (anchor !== null && head !== null) {
-      const selection = relSel.isNodeSelection
+      const selection = relSel.type === 'NodeSelection'
         ? NodeSelection.create(tr.doc, Math.min(anchor, head))
         : TextSelection.create(tr.doc, anchor, head)
       tr = tr.setSelection(selection)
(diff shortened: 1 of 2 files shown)


src/plugins/sync-plugin.js
```diff
@@ -194,14 +194,18 @@ const restoreRelativeSelection = (tr, relSel, binding) => {
...
+      tr = tr.setSelection(selection)
     }
   }
 }
 
 export const getRelativeSelection = (pmbinding, state) => ({
   anchor: absolutePositionToRelativePosition(state.selection.anchor, pmbinding.type, pmbinding.mapping),
-  head: absolutePositionToRelativePosition(state.selection.head, pmbinding.type, pmbinding.mapping)
+  head: absolutePositionToRelativePosition(state.selection.head, pmbinding.type, pmbinding.mapping),
+  isNodeSelection: state.selection instanceof NodeSelection,
```
dmonad, line 208: Can you please change this to `type: 'NodeSelection' | 'TextSelection'`? This would make it possible to potentially add more types of selections.

src/plugins/sync-plugin.js
```diff
@@ -194,14 +194,18 @@ const restoreRelativeSelection = (tr, relSel, binding) => {
     const anchor = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.anchor, binding.mapping)
     const head = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.head, binding.mapping)
     if (anchor !== null && head !== null) {
-      tr = tr.setSelection(TextSelection.create(tr.doc, anchor, head))
+      const selection = relSel.isNodeSelection
+        ? NodeSelection.create(tr.doc, Math.min(anchor, head))
```
dmonad, line 198: How does this behave when multiple nodes are selected?

dmonad (changes requested):
Could you please add a test-case for this?

src/plugins/sync-plugin.js
jamesopti: > How does this behave when multiple nodes are selected?

Hmmm good question! I'm actually not sure how to test this.

jamesopti: According to the Prosemirror docs, a [Node Selection](https://prosemirror.net/docs/ref/#state.NodeSelection) _is a selection that points at a single node._ 

I believe a range that points to mulitple nodes or ranges is still considered a Text Selection, however, I don't know how or if it is possible to select multiple specific nodes at once.

jamesopti:
@dmonad I think the complete solution here to handle custom selections (like Cell Selection for tables) is to accept an optional function parameter `restoreSelection(tr, relSel)` that a consumer could implement to handle their own selection restoration.

If you think that makes sense to do now, I'm happy to do it as part of this PR or as a follow up.

E.g.

```ts
const defaultRestoreSelection = (tr, relSel) => {
  const selection = relSel.type === 'node'
    ? NodeSelection.create(tr.doc, Math.min(anchor, head))
    : TextSelection.create(tr.doc, anchor, head)
  tr = tr.setSelection(selection)
}


export const ySyncPlugin = (yXmlFragment, {
  colors = defaultColors,
  colorMapping = new Map(),
  permanentUserData = null,
  onFirstRender = () => {},
  restoreSelection = defaultRestoreSelection
} = {}) => {

  ...

})

const restoreRelativeSelection = (tr, relSel, binding, restoreSelection) => {
  if (relSel !== null && relSel.anchor !== null && relSel.head !== null) {
    const anchor = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.anchor, binding.mapping)
    const head = relativePositionToAbsolutePosition(binding.doc, binding.type, relSel.head, binding.mapping)
    if (anchor !== null && head !== null) {
      restoreSelection(tr, relSel)
    }
  }
}
```

jamesopti:
@dmonad Any chance we could get this merged soon?

dmonad:
FYI this is pretty low on my priority list. I just took some time off. Now I need to catch up on quite a lot of stuff. The good thing is that it is on some kind of priority list I guess :sweat_smile:

amilich:
> FYI this is pretty low on my priority list. I just took some time off. Now I need to catch up on quite a lot of stuff. The good thing is that it is on some kind of priority list I guess 😅

Hey all! Just noting that we could use this fix as well. Would help us out fixing some other selection types.

amilich:
> > FYI this is pretty low on my priority list. I just took some time off. Now I need to catch up on quite a lot of stuff. The good thing is that it is on some kind of priority list I guess 😅
> 
> Hey all! Just noting that we could use this fix as well. Would help us out fixing some other selection types.

Checking in briefly on this

rebolyte:
Just bumped into this as well. @jamesopti anything left before it can be merged?

Alecyrus:
I have same issue https://github.com/yjs/y-prosemirror/issues/43. I try this commit and it does not work, but https://github.com/hamflx/y-prosemirror/commit/2ed27a029e47f292b388ac950a3fbdf5fdf8f30b works.

nperez0111:
Superceded by: https://github.com/yjs/y-prosemirror/pull/181

Issue #126, opened 2022-09-19
Y-ProseMirror always set the caret at the end of the document after initialize the editor


ocavue:
**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**

I want to initialize the editor with a particular selection position (e.g. at the start of the document), but y-prosemirror always changes the initial selection to the end of the document. 

ProseMirror allows me to set the initial selection by providing an initial editor state with a selection. The code is shown below: 

```ts
  const prosemirrorView = new EditorView(editor, {
    state: EditorState.create({
      doc,
      selection, // <-- here is my initial selection
      schema,
      plugins: [...]
    })
  })
```

Even if the initial selection is not implicitly passed, the default selection is at the start of the document. 

When I use `ySyncPlugin`, `_forceRerender` will replace the content of the whole document, which make the selection at the end of the document. 

**To Reproduce**

I prepare a simple example based on the `demo/` example in this repo. You can find the code [here](https://github.com/ocavue/y-prosemirror/pull/1/files). 

```
$ git clone https://github.com/ocavue/y-prosemirror -b ocavue/init_selection
$ cd y-prosemirror
$ npm install 
$ npm run start 
```

Go to http://127.0.0.1:8081/demo/prosemirror.html. The editor should be focused and the caret is at the end of the document. 

If `ENABLE_YJS` in `demo/prosemirror.js` is set as `false`, then the `y-prosemirror` in the heading will be selected after a reload.

Notice that in this example, to set the initial selection, I have to pass the initial document with two different ways in the same time:
- pass the `doc` to `EditorState`.
- pass the `yDoc` to `ySyncPlugin`.

**Expected behavior**

I can set the initial selection when using y-prosemirror. 

**Screenshots**

https://user-images.githubusercontent.com/24715727/191005843-46a93d9c-9e5e-4ffc-85cc-4f46a60a0294.mp4

**Environment Information**
 - Browser: Chrome version 105
 - Yjs version and the versions of the y-* modules you are using
   - y-prosemirror@1.1.3 
   - yjs@13.5.38

**Additional context**

I created a PR https://github.com/yjs/y-prosemirror/pull/127 to resolve this issue.

HondryTravis:
same question

[...]

PR #127, opened 2022-09-19
skip unnecessary full-document replaces

ocavue:
This PR skips some unnecessary replaces by comparing the incoming content and current content in the ProseMirror editor view. This could be a potential solution for https://github.com/yjs/y-prosemirror/issues/126

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -375,11 +375,14 @@ export class ProsemirrorBinding {
           this.mapping
         )
       ).filter((n) => n !== null)
-      // @ts-ignore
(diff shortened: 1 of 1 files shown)


maccman:
cc @dmonad :)

maccman:
Ping :)

martinboksa:
ping :)

fuadsaud:
Has anyone verified this to fix #126? I rebase this PR on master and tried pointing my app to it, but I still see the same behavior (it's not impossible that I'm doing something wrong when building the library locally, given that I depend on y-prosemirror indirectly through tiptap).

PR #129, opened 2022-11-21
move typescript to devDependencies

ocavue:
`y-prosemirror` doesn't depend on `typescript` during runtime, so I guess `typescript` should be moved to `devDependencies`.


dmonad:
Absolutely, thanks!

PR #134, opened 2023-04-10
Allow skipping YDoc updates for transactions

elad-barack:
YJS integration with our editor encountered a problem. We have some PM nodes & marks that are per-session. We do not want to update other sessions about their existence and changes when they are in the doc. 

- Some of them states that the user is in a middle of an "inner" session, and usually there's a floating card displayed when they are on the screen. 
- The other are kind of a "dictionary" marks. We don't want the other session to receive the same mark, even if it sound like we should on the first glance, since other parts of our editor screen won't be synced with it, which is problematic for our users. 

We just used `addToHistory` flag until we added YJS. It worked great :) 
The change is basically a very small one. We added a state param called `addToYDoc` that is true by default, and false only when a user passes `addToYDoc` in the transaction meta. When it's false, the change won't be reflected in the doc history, so we won't update other sessions about the redundant change.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -113,6 +113,7 @@ export const ySyncPlugin = (yXmlFragment, {
           isChangeOrigin: false,
           isUndoRedoOperation: false,
           addToHistory: true,
+          addToYDoc: true,
           colors,
           colorMapping,
           permanentUserData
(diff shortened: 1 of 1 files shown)


elad-barack:
@dmonad Is there any chance someone will take a look and merge? 
Many people reacted to the PR, so it might help others as well..

sradu:
Would be great to have it!

dmonad:
Maybe I don't understand this feature, but wouldn't the changes be merged into Yjs the next time a change is emitted? Have you been able to test that this works?

In any case, I feel there is a better solution to this problem.

It would be nice to see Yjs in nextcloud. Could you explain in more detail what you are trying to achieve?

Issue #47, opened 2021-05-03
Does not support marks on non-TextNodes


ascott18:
**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**

This plugin fundamentally does not support mark on nodes that are not text nodes. For example, Tiptap's image node, which is an inline, non-text node: https://github.com/ueberdosis/tiptap/blob/v1/packages/tiptap-extensions/src/nodes/Image.js.

**To Reproduce**
1. Apply a mark to an inline node. For example, apply a hyperlink (https://github.com/ueberdosis/tiptap/blob/v1/packages/tiptap-extensions/src/marks/Link.js) to an image (https://github.com/ueberdosis/tiptap/blob/v1/packages/tiptap-extensions/src/nodes/Image.js)
2. Be the editor of the document. Observe that the JSON version of the prosemirror doc does contain the expected mark, and that the HTML is rendered with the correct `<a></a>` wrapping the image node.
2. Be a realtime collaborator on the same document, OR reload the document using the serialized ydoc. Observe that the hyperlink on the image is missing because the mark is not serialized into the ydoc.

**Expected behavior**

Marks on custom nodes are handled just as attributes on custom nodes are handled.

BrianHung:
If anyone wants to tackle this in a PR, two main functions to address are `createTypeFromElementNode` and `createNodeFromYElement`. 

Particularly, serialize the marks of a node as `Y.XmlElement` attributes using `setAttribute` while differentiating them somehow from `node.attrs`, then deserializing them and passing `Mark[]` as argument to [`schema.node`](https://prosemirror.net/docs/ref/#model.Schema.node).

alexharri:
The lack of support for marks on custom nodes is currently causing issues in an editor that I'm building. Would love to see this feature.

Judging from the title, #63 sounds like it should resolve the issue. Is there something preventing us from taking #63 in @dmonad? I would love to help if I can.

rebolyte:
Same here, this has become a blocker on one of my projects as well. Can #63 be merged?

martinboksa:
Any updates, please? @dmonad

icode:
The same problem, custom nodes cannot support marks

pointnet:
Same problem here, any updates?

[...]

PR #135, opened 2023-04-24
fix: unsupported node mark

XiaodongTong:
fix https://github.com/yjs/y-prosemirror/issues/47

Using『node mark』 as an attribute on YElement，Each mark corresponds to an YElement attribute.  
Runable in my project. 

Hope to merge

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -671,7 +671,19 @@ const createNodeFromYElement = (
           : { type: 'added' }
       }
     }
-    const node = schema.node(el.nodeName, attrs, children)
+    const nodeAttrs = {};
+    const nodeMarks = [];
+
+    for (const key in attrs) {
+      if (key.includes('yelement_mark_')) {
+        const markName = key.replace('yelement_mark_','');
+        let markValue = JSON.parse(attrs[key]);
+        nodeMarks.push(schema.mark(markName, markValue))
+      } else {
+        nodeAttrs[key] = attrs[key]
+      }
+    }
+    const node = schema.node(el.nodeName, nodeAttrs, children, nodeMarks);
     mapping.set(el, node)
     return node
   } catch (e) {
@@ -750,11 +762,15 @@ const createTypeFromTextNodes = (nodes, mapping) => {
  */
 const createTypeFromElementNode = (node, mapping) => {
   const type = new Y.XmlElement(node.type.name)
+  const nodeMarksAttr = nodeMarksToAttributes(node.marks);
   for (const key in node.attrs) {
     const val = node.attrs[key]
     if (val !== null && key !== 'ychange') {
       type.setAttribute(key, val)
     }
+  }
+  for (const key in nodeMarksAttr) {
+    type.setAttribute(key, nodeMarksAttr[key])
   }
   type.insert(
     0,
@@ -965,6 +981,17 @@ const marksToAttributes = (marks) => {
   return pattrs
 }
 
+const nodeMarksToAttributes = (marks) => {
+  const pattrs = {}
+  const prefix = 'yelement_mark_'
+  marks.forEach((mark) => {
+    if (mark.type.name !== 'ychange') {
+      pattrs[`${prefix}${mark.type.name}`] = JSON.stringify(mark.attrs || {})
+    }
+  })
+  return pattrs
+}
+
 /**
  * @private
  * @param {{transact: Function}} y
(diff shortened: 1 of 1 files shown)


src/plugins/sync-plugin.js
```diff
@@ -965,6 +1001,17 @@ const marksToAttributes = (marks) => {
   return pattrs
 }
 
+const nodeMarksToAttributes = (marks) => {
+  const pattrs = {}
+  const prefix = 'yelement_mark_'
+  marks.forEach((mark) => {
+    if (mark.type.name !== 'ychange') {
+      pattrs[`${prefix}${mark.type.name}`] = JSON.stringify(mark.attrs || {})
```
dmonad, line 1009: Why Stringify?

src/plugins/sync-plugin.js
```diff
@@ -984,18 +1031,21 @@ export const updateYFragment = (y, yDomFragment, pNode, mapping) => {
...
+          yDomFragment.setAttribute(key, attrs[key])
         }
       } else {
         yDomFragment.removeAttribute(key)
       }
     }
     // remove all keys that are no longer in pAttrs
     for (const key in yDomAttrs) {
-      if (pAttrs[key] === undefined) {
+      if (attrs[key] === undefined) {
```
dmonad, line 1048: Here, we clean up unused attributes. When are unused marks removed?

src/plugins/sync-plugin.js
```diff
@@ -794,6 +810,25 @@ const equalAttrs = (pattrs, yattrs) => {
...
+
+  let pMarkAttr = nodeMarksToAttributes(pmarks);
+
+  console.log('yattrs:',yattrs , 'pMarkAttr',pMarkAttr);
+
+  for (let i = 0; i < keys.length && eq; i++) {
+    const key = keys[i];
+    const l = JSON.stringify(pMarkAttr[key])
+    const r = yattrs[key]
+    eq = key === 'ychange' || l === r ;
```
dmonad, line 827: Once `l` and `r` are objects. You could use `f.equalityDeep` to compare them. Where `import * as f from 'lib0/function'`

dmonad (changes requested):
Thanks for the PR. Could you please add some tests first? Specifically, we should test adding & removing marks from nodes.

src/plugins/sync-plugin.js
XiaodongTong: i will change

src/plugins/sync-plugin.js
```diff
@@ -671,7 +672,19 @@ const createNodeFromYElement = (
...
       }
     }
-    const node = schema.node(el.nodeName, attrs, children)
+    const nodeAttrs = {};
+    const nodeMarks = [];
+
+    for (const key in attrs) {
+      if (key.startsWith('yelement_mark_')) {
+        const markName = key.replace('yelement_mark_','');
+        let markValue = attrs[key];
```
dmonad, line 681: should be `const`

src/plugins/sync-plugin.js
```diff
@@ -780,7 +797,7 @@ const createTypeFromTextOrElementNode = (node, mapping) =>
 const isObject = (val) => typeof val === 'object' && val !== null
 
 const equalAttrs = (pattrs, yattrs) => {
-  const keys = Object.keys(pattrs).filter((key) => pattrs[key] !== null)
+  const keys = Object.keys(pattrs).filter((key) => pattrs[key] !== null && !key.startsWith('yelement_mark_'))
```
dmonad, line 800: Why are you filtering `yelement_mark_` here in the attributes that you get from the prosemirror node? If there is a reason, why are you not filtering in the next line when you are comparing?

src/plugins/sync-plugin.js
```diff
@@ -965,6 +1000,17 @@ const marksToAttributes = (marks) => {
   return pattrs
 }
 
+const nodeMarksToAttributes = (marks) => {
+  const pattrs = {}
+  const prefix = 'yelement_mark_'
```
dmonad, line 1005: Let's store this prefix globally somewhere as it is reused throughout the file. Also I suggest we use a shorter name that starts with an `_`. E.g. `_mark_`.

dmonad (changes requested):

josonyang:
> I meet similar sitituation in my custom prosemirror's schema.I'm not sure about when to use mark / attrs for a node.
> 
> But in file y-prosemirror/src/lib.js， yXmlFragmentToProsemirrorJSON this function processes only text nodes' marks, and attrs only for non-text nodes.
> 
> If marks are designed just for text node, and atts for non-text nodes, y-prosemirror will should not support marks for non-text nodes, as that will be anti prosemirror's design.
> 
> It's a pleasure for any suggestion or guidance on how to use marks/attrs correctly, fellows the design.

I have also need to handle this case

In my understanding，marks act as a way to add attrs to prosemirror document , not only text node.

In tiptap and Confluence Editor , also use mark to effect non-text node

jjgmckenzie:
Looking good, very useful feature for format compatibility & avoiding data loss.

@XiaodongTong - any plans to fix the final requested change?

hoebbelsB:
@dmonad are there any plans to merge this? I also face the very same issue and wanted to raise a PR. Then I noticed this one here. I can also complete this work if you like

Pruxis:
What's actuality pending here? I'd like to contribute to get this thing through

PR #137, opened 2023-07-03
Fix to load type definition in esm

kobayang:
I'm trying to y-prosemirror with TypeScript `moduleResolution: "Bundler"` mode.
But TypeScript can not find y-prosemirror type definition in this mode.

To import the types correctly, we need to add `exports.types` field.
https://www.typescriptlang.org/docs/handbook/esm-node.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   },
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


dmonad:
Thanks for closing this ticket. I hope the issue has been resolved!

PR #140, opened 2023-09-06
feat: Added `awarenessStateFilter` option for filtering `awareness.getState()`

RubaXa:
This option is needed when we want to remove self "cursor" when opened N-tabs or other logic.

For example, in "GoogleDocs", this is exactly the behavior, they display N-1 "cursors" (N — number of tabs), but in the case of "Notion", they do not display their own "cursor".

```js
yCursorPlugin(provider.awareness, {
  awarenessStateFilter: (_, __, user) => user.id !== getAuthUserId(),
})
```

### DIFF
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -12,6 +12,16 @@ import { yCursorPluginKey, ySyncPluginKey } from './keys.js'
 import * as math from 'lib0/math'
 
 /**
+ * Default awareness state filter
+ *
+ * @param {number} currentClientId current client id
+ * @param {number} userClientId user client id
+ * @param {any} user user data
+ * @return {boolean}
+ */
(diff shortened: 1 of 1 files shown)


dmonad:
Please explain the usecase first. Have you considered other solutions?

Also, this doesn't really prevent unauthorized users from sending awareness states. Instead, you should filter messages from unauthorized users.

RubaXa:
@dmonad If I open a document in 2-tabs, I will see "cursor" from myself.

For example, in "GoogleDocs", this is exactly the behavior, they display N-1 "cursors" (N — number of tabs), but in the case of "Notion", they do not display their own "cursor".

Therefore, I want to be able to regulate this behavior, for example, through filtering.

dmonad:
Okay, makes sense. Will get this into the next release.

eldh:
Any chance we'll see a new release soon? Just found myself needing this…

PR #142, opened 2023-09-13
Inspect meta of `appendedTransaction` for possible `addToHistory:false

### LINKED ISSUE #141: `ySyncPlugin` ignores `addToHistory:false` meta of `appendedTransaction`
**Describe the bug**
`ySyncPlugin` ignores `addToHistory: false` of `appendedTransaction` thus root transaction gets added onto undo/redo stack.

**To Reproduce**
1. Clone fork with a test repro: https://github.com/romansp/y-prosemirror/tree/appendedTransaction-addToHistory-false
2. Run `npm run test`
3. See that test named `add to history ignore with append transaction plugin` is failing

**Expected behavior**
`addToHistory: false` should be respected if it comes from `appendedTransaction`. Root transaction should not appear on the undo/redo stack, and undoing/redoing should not have any affect.

**Actual behavior**
Transaction is added onto undo stack.

**Environment Information**
 - Browser, Microsoft Edge 116.0.1938.76 (Official build) (arm64)
 - `y-prosemirror` v1.2.1
 - `yjs` v13.6.7

**Additional context**
Original `prosemirror-history` plugin does indeed checks `appendedTransaction` and inspects for possible `addToHistory: false` as demonstrated here https://github.com/ProseMirror/prosemirror-history/blob/40d274a74d0fc0787aeca03634a64d0c78f18a50/src/history.ts#L277C15-L277C15

romansp:
Resolves #141.

Using the same approach as in native `prosemirror-history`, added inspection of `appendedTransaction` meta for possible `addToHistory: false` to prevent root transaction from appearing on the undo/redo stack.

Added a test case that covers expected behavior.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -126,7 +126,8 @@ export const ySyncPlugin = (yXmlFragment, {
             pluginState[key] = change[key]
           }
         }
-        pluginState.addToHistory = tr.getMeta('addToHistory') !== false
+        const appended = tr.getMeta('appendedTransaction')
+        pluginState.addToHistory = tr.getMeta('addToHistory') !== false && !(appended && appended.getMeta('addToHistory') === false)
         // always set isChangeOrigin. If undefined, this is not change origin.
         pluginState.isChangeOrigin = change !== undefined &&
           !!change.isChangeOrigin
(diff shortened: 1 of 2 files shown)


BrianHung:
Great fix for developers trying to chain multiple commands together but only have one undo / redo for all using the `appendedTransaction` meta.

Issue #85, opened 2021-12-10
Infinite cursor awareness update loop when two prosemirror docs edit same underlying y.XmlFragment


milesingrams:
**Describe the bug**
I am using TipTap (https://tiptap.dev/) which uses y-prosemirror under the hood to handle collaboration cursors. I have a semi unique situation where I have a page with two editors both which edit the same Y.XmlFragment. When I click on either of the editors to begin editing the cursor starts flickering and runs and infinite cycle of updating the awareness state. I have narrowed down the cycle to this block of code:

https://github.com/yjs/y-prosemirror/blob/984175f7fd7960b1478f10370829659f001ccd3b/src/plugins/cursor-plugin.js#L113-L136

The infinite loop is cycling between:

```javascript
awareness.setLocalStateField(cursorStateField, { 
  anchor, head 
}) 
```

and 

```javascript
awareness.setLocalStateField(cursorStateField, null)
```

My guess is that while the cursor plugin beautifully handles multiple prosemirror documents on a page, it has issues when two prosemirror documents edit the same underlying Y.XmlFragment.

Perhaps more logic around the editor gaining and losing focus could resolve this. i.e. only update awareness when editor is actively focused and set awareness to null when editor loses focus. 

**Expected behavior**
I expect multiple editors to be able to seamlessly edit the same Y.XmlFragment without causing the awareness updates to enter an infinite loop.

**Environment Information**
 - Browser: Chrome
 - Yjs: "^13.5.22"
 - y-webrtc: "^10.2.2"
 - y-prosemirror: "^1.0.13"

Anyways I'm sure this isn't the most common use case but I would greatly appreciate the help. Thank you for such an awesome tool!

milesingrams:
I managed to get this to work by explicitly unsetting the cursor state only on `focusout`. Not sure if this code will inevitably cause issues but seems to be working pretty smoothly so far.

Here is the modified code I am using:

```javascript
const updateCursorInfo = () => {
	const ystate = ySyncPluginKey.getState(view.state)
	const current = awareness.getLocalState() || {}
	if (view.hasFocus() && ystate.binding) {
		const selection = getSelection(view.state)
		const anchor = absolutePositionToRelativePosition(selection.anchor, ystate.type, ystate.binding.mapping)
		const head = absolutePositionToRelativePosition(selection.head, ystate.type, ystate.binding.mapping)
		if (
			!current[cursorStateField] ||
[...]

PR #145, opened 2023-10-12
Prevent infinite cursor awareness update loop

furkan3ayraktar:
If two editors are open for the same Y.Doc, the cursor awareness state enters into an infinite update loop. This commit fixes the issue by using the solution suggested in [this issue](https://github.com/yjs/y-prosemirror/issues/85).

We've been using this solution for the last one and half years in production without any issues. I remembered it when receiving a recent notification from the issue and wanted to share the solution as a PR.

### DIFF
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -241,27 +241,26 @@ export const yCursorPlugin = (
               head
             })
           }
-        } else if (
-          current.cursor != null &&
-          relativePositionToAbsolutePosition(
-            ystate.doc,
-            ystate.type,
-            Y.createRelativePositionFromJSON(current.cursor.anchor),
-            ystate.binding.mapping
-          ) !== null
-        ) {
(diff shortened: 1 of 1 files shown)


szymonkozak:
@furkan3ayraktar Should it also help in the React strict mode? I tried to use the patch and it doesn't help when Strict mode is enabled (there is still a problem with the cursor infinite update loop).

szymonkozak:
I have tested this solution and though it fixes the main problem I have noticed that after switching a tab (`focusout` event), the cursor of the user (which switches the tab) is blinking for a moment. When we don't register `focusout` listener then this problem is solved but I don't know if it introduces other issues or not.

garth:
I used pnpm patch to apply this PR and it fixed infinite loops when using awareness with tiptap.

euvl:
Hey there,
Hocuspocus seem to use y-prosemirror as well, any idea how i can apply this patch in there? 

Thanks

PR #148, opened 2023-11-06
fix: remove random number in cursor selection builder

baotpham:
this causes selection color not rendering properly


icode:
I don't understand why we need to add 70

troglodytto:
This is to make sure that the Alpha Channel of the Color is set to 70% opacity. Hence, when selecting the text, it overlays and does not hide the content below it. However, it only supports Hex Color as you can see in the code, and **PR https://github.com/yjs/y-prosemirror/pull/154** fixes this for HSLA/RGB color codes

PR #151, opened 2023-12-21
Update peer dependencies package.json

websiddu:


hoebbelsB:
why isn't this merged? prosemirror-model 1.7 was released 5 years ago. The most recent version is `1.2.0`. Because they messed up their versioning, we are always facing the following warning

image

This PR would fix that issue :)

dmonad:
I'm very happy to fix warning messages. But I didn't know about this one. Should I? Who throws it?

I don't understand why I should update peer-dependencies if I didn't break compatibility with the mentioned packages.

y-prosemirror is still compatible with prosemirror-model `1.7.0` (yes, the one that was released 5 years ago). But y-prosemirror is also compatible with prosemirror-model `1.20.0` (the latest release). `^1.7.0` should match `1.20.0` (semantic versioning).

dmonad:
> why isn't this merged? prosemirror-model 1.7 was released 5 years ago. The most recent version is `1.2.0`. Because they messed up their versioning, we are always facing the following warning
> 
> image
> 
> This PR would fix that issue :)

Okay. I think you misunderstand who causes this issue. The warning message literally says that ***you*** are using an old version of prosemirror-model.

The latest release of prosemirror-model is `1.20.0` (not 1.2.0, that one is even older). 

I don't see a need to update my peer-dependencies.

hoebbelsB:
@dmonad thanks for pointing that one out, sorry for the confusion 🤦

dmonad:
Np, that happens :wink:

PR #154, opened 2024-05-31
feat: implement generic color capabilities

troglodytto:
- implement color utils to allow the usage of multiple kinds of colors instead of *JUST* RGB
- now supports Hex, RGB, HSL
- includes test cases for 3 different formats of colors

### DIFF
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -10,6 +10,7 @@ import {
 import { yCursorPluginKey, ySyncPluginKey } from './keys.js'
 
 import * as math from 'lib0/math'
+import { setAlphaChannel } from 'src/utils/colors.js';
 
 /**
  * Default awareness state filter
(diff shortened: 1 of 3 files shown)


troglodytto:
@dmonad Just bringing this to attention

piyush-kroolo:
I too would like to see this @dmonad. Right now, I'm having to use a hex color, but I'd like to use an HSL value so that I can decide the brightness of the colors depending on the theme that the user is on, and this feature would make it easier to implement without having to fall back to first converting the HSL values to Hex (by either implementing the conversion code ourselves or relying on third party libraries like tinycolor etc.) and then using that hex as the input.

troglodytto:
Thanks for your input @dmonad . 

Yes, We can keep this PR open for now.

Can you please let me know which kind of potential errors other extensions might throw?

I've tried to keep this as backwards compatible as possible. 

But I'll try finding solutions to make it fully compatible with other extensions in the meantime, basis on the potential errors that you mention.

PS. For Anyone else reading this, an alternative approach would be to use something like https://github.com/bgrins/TinyColor and convert your colors to a Hex string (specifically a 6 digit Hex string, Not a 3,4 or 8 digit one)

dmonad:
It's just that other extensions don't understand hsl. Supporting hsl is like opening Pandora's box: eventually we'd need to support all formats in all extensions. These formats basically do the same thing, so I'm not sure I want to open that box.

PR #157, opened 2024-07-02
support marks on element nodes

Pruxis:
https://github.com/yjs/y-prosemirror/pull/63
https://github.com/yjs/y-prosemirror/pull/135

But updated so that it works to parse yDoc to JSON aswell.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1,4 +1,4 @@
-import { updateYFragment, createNodeFromYElement } from './plugins/sync-plugin.js' // eslint-disable-line
+import { updateYFragment, createNodeFromYElement, MarkPrefix } from './plugins/sync-plugin.js' // eslint-disable-line
 import { ySyncPluginKey } from './plugins/keys.js'
 import * as Y from 'yjs'
 import { EditorView } from 'prosemirror-view' // eslint-disable-line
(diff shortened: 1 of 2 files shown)


nperez0111:
Hi @dmonad, as I understand it you wanted to add some tests for this. I had a hand in making this change and am willing to add tests if you find it necessary.

Pruxis:
Anything pending to get this merged?

aoruize:
Please get this merged 🙏🏼 🙏🏼 😭 😮‍💨

arhamfhq:
Can somebody share any update we really need this issue fixed asap.

nperez0111:
I've added a couple of tests for this to prove that it works as intended.

The add test fails before this change, the remove test passes because the mark was never actually preserved in the ydoc

dmonad:
I'm sorry that I've been sitting on this for so long.

This PR makes quite a few changes that I can't take back easily. I know it fixes your immediate use-case. But longer-term, I believe that we should always go for the right solution.

I'm not happy with using Yjs attributes to sync marks. y-prosemirror reflects marks as formatting attributes. Hence, I believe the correct approach would be to use Yjs formatting attributes instead of attributes to sync marks. 

There might be the argument that attributes and formatting attributes are semantically the same thing. If that were true, you could use prosemirror attributes instead of marks. 

All I can say right now is that there is more to research here.

nperez0111:
@dmonad do YXMLFragments support Yjs formatting attributes?

Do you think it would be difficult to implement in that way?

dmonad:
@nperez0111  Unfortunately, not yet. But I might be able to work on this next year.

nperez0111:
@dmonad no worries!

If you need help at all, I'm happy to help. I've dived into the yjs codebase a couple of times now for things I needed. So maybe I can help move this forward somehow. I could dig around to understand the current implementation, but even a high-level description of how you'd see it working would be helpful

jensug:
Hi, any updates on this? :)

PR #158, opened 2024-07-09
Removed reference to travis in readme

tomasdahlqvist:
Version 1.0.17 removed .travis.yml and now the build status is broken.
If travis is reimplemented then the link is to https://app.travis-ci.com/github/yjs/y-prosemirror


dmonad:
Right, thanks!

PR #159, opened 2024-07-11
Fix `cursorStateField` bug

### LINKED ISSUE #86: Update current.cursor references to current[cursorStateField]?
Not sure if this was intended, but since the `cursorStateField` is customizable I figured the references to `current.cursor` should be changed to `current[cursorStateField]`. I may not understand the code fully though so apologies if this is there for a reason.

https://github.com/yjs/y-prosemirror/blob/984175f7fd7960b1478f10370829659f001ccd3b/src/plugins/cursor-plugin.js#L127-L132

andrictham:
# Summary
This PR fixes a long-standing issue with the `cursorStateField` param that’s passed into `yCursorPlugin`.

Setting `cursorStateField` is the only way to namespace awareness state so that cursors from different subdocuments syncing over the same provider will be able to read and write their own cursor state.

However, it is currently broken. Setting `cursorStateField` doesn’t work as intended: editors write cursor state to the correct field, but doesn’t read back from it when rendering cursors.

This fixes #86

# Current Behavior
If a custom value of `cursorStateField` is set, the cursor plugin correctly uses it to set awareness state: 
https://github.com/yjs/y-prosemirror/blob/fd08cf3c7c307c1446178765baa94603a3a8416a/src/plugins/cursor-plugin.js#L239-L242

However, in the `createDecorations` function, the value `aw.cursor` is still used. This is reading a hardcoded field with the name `"cursor"` from the awareness state. 
https://github.com/yjs/y-prosemirror/blob/fd08cf3c7c307c1446178765baa94603a3a8416a/src/plugins/cursor-plugin.js#L101-L112

This results in cursors not rendering properly if you set `cursorStateField` to a custom value.

# Proposed Changes

- All references to a hardcoded `"cursor"` field are replaced with a dynamic reference to the field name set in `cursorStateField`.

- In `updateCursorInfo`, the condition to clear cursor state and set it to `null` is simplified, so that when the `focusout` event fires, we always run `awareness.setLocalStateField(cursorStateField, null)`. Without this change, if the editor loses focus, the cursor state isn’t properly cleared, which leads to lingering cursors.

# Tests
- I ran `npm run test` and confirmed that all tests are successful
- I manually tested this in my own repo and confirmed that this fixes the original issue where cursor state between subdocuments are wrongly replicated. With this fix, if a subdocument’s guid is passed into the `cursorStateField` param, each subdocument’s cursor state will now sync to separate awareness fields.

### DIFF
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -186,7 +186,8 @@ export const yCursorPlugin = (
             awareness,
             awarenessStateFilter,
             cursorBuilder,
-            selectionBuilder
+            selectionBuilder,
+            cursorStateField
           )
         }
         return prevState.map(tr.mapping, tr.doc)
--- a/src/plugins/cursor-plugin.js
+++ b/src/plugins/cursor-plugin.js
@@ -87,7 +87,6 @@ export const createDecorations = (
   awareness.getStates().forEach((aw, clientId) => {
     if (!awarenessFilter(y.clientID, clientId, aw)) {
       return
-      return;
     }
 
     if (aw[cursorStateField] != null) {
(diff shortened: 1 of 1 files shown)


andrictham:
@dmonad Let me know once you’ve had a look at this! 🙏

andrictham:
@dmonad I wonder if it’s possible to get this merged, or if not, is there something I’m misunderstanding about how `cursorStateField` is supposed to work, or something I'm missing in my PR? 

I’m reliant on this fix so I can get subdocuments working in my app. 

Let me know how I can help 😀

Issue #170, opened 2024-11-21
Out of range position exception when adding ySyncPlugin to an existing editor


gvergnaud:
## **Describe the bug**

When dynamically adding the `ySyncPlugin` plugin to an existing editor with a state that doesn't perfectly match the current the `Y.XmlFragment` state, the ySyncPlugin overrides the editor state to fix this inconsistency (which is expected), and then tries to preserve the user's current selection ([here in the code](https://github.com/yjs/y-prosemirror/blob/876695b2d630f6de1becf87ff2d8401b36c24380/src/plugins/sync-plugin.js#L411-L413)).

This second part **can throw an exception** if the position is no longer valid for the updated document state.

<img width="626" alt="Screenshot 2024-11-21 at 14 41 32" src="https://github.com/user-attachments/assets/b0b9f315-2a6a-4a94-a573-8918feeb6fe9">

## **To Reproduce**

You can reproduce this error in [this stackblitz sandbox](https://stackblitz.com/edit/vitejs-vite-7xewkn?file=src%2FApp.tsx,src%2FApp.css,src%2Findex.css,src%2Fmain.tsx&terminal=dev).

0. Open [the link to the reproduction case](https://stackblitz.com/edit/vitejs-vite-7xewkn?file=src%2FApp.tsx,src%2FApp.css,src%2Findex.css,src%2Fmain.tsx&terminal=dev).
1. Start by typing in the editor.
2. Click on the button. It will add the ySyncPlugin to the editor's list of plugins.
3. An out-of-bounds error will be thrown.
4. Open the console to look at the error.

## **Expected behavior**

When applying your previous selection to the new state, ySyncPlugin should clamp it to the range of valid positions of the new document state, to avoid throwing an error.

**Environment Information**
 - "y-prosemirror": "1.2.12"
 -  "yjs": "13.6.20"

PR #171, opened 2024-11-21
fix: avoid throwing an exception if the current selection no longer makes sense when initializing ySyncPlugin

gvergnaud:
Fixes https://github.com/yjs/y-prosemirror/issues/170.


dmonad:
Awesome PR!

- Precise bug report
- Reproducible code
- Working fix that even uses lib0.math instead of `Math`. 

It took me just a few minutes and saved me a lot of time. Thanks!!

Will merge, and release now. There should be a new release with a fix in a few minutes.

PR #172, opened 2024-11-22
feat: support overlapping marks #34 based on (#52)

nperez0111:
This adds support for overlapping marks to both text elements and nodes, it is based on (#52) but is:
 - based in a much more recent branch which adds node mark support (#157)
 - supports multiple node marks based on the work set up by (#157)
 - passes tests


mdoesburg:
@dmonad Would love for this to be supported! Is this something you'd consider merging?

bdbch:
Great work!

PR #176, opened 2025-02-09
fix: avoid throwing an exception if the restored selection does not make sense

vitmf:
Similarly to https://github.com/yjs/y-prosemirror/issues/170, in some rare cases `restoreRelativeSelection` in `_typeChanged` can't restore the selection and throws an exception. This happens when the yjs document can't be perfectly converted to the prosemirror document because of schema restrictions.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -245,7 +245,9 @@ const restoreRelativeSelection = (tr, relSel, binding) => {
       binding.mapping
     )
     if (anchor !== null && head !== null) {
-      tr = tr.setSelection(TextSelection.create(tr.doc, anchor, head))
+      const clampedAnchor = math.min(math.max(anchor, 0), tr.doc.content.size)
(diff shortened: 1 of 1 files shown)


nperez0111:
I think is stale by: https://github.com/yjs/y-prosemirror/pull/176

vitmf:
https://github.com/yjs/y-prosemirror/pull/181 (I think it is the right PR) still does not solve the issue. The exception is thrown by `tr.doc.resolve`, and not by `TextSelection.create`.

nperez0111:
> #181 (I think it is the right PR) still does not solve the issue. The exception is thrown by `tr.doc.resolve`, and not by `TextSelection.create`.
> 
> Edit: [here](https://stackblitz.com/edit/vitejs-vite-jxfgzmld) is a very contrived example.

You are right, this is still an issue, I made a PR based on the newer version: https://github.com/yjs/y-prosemirror/pull/185

PR #180, opened 2025-04-09
fix: pass historyDoc type in renderSnapshot

Ft-Blue:
This fixes an issue where the document type used for creating the document fragment to render snapshots is not updated with `historyDocument``

**Observed issue:** When the initial document has garbage collection enabled, deleted content **_is GC'ed_** in `historyDoc` when rendering the diff between 2 snapshots.

**Expected behavior:** When the initial document has garbage collection enabled, deleted content _**is kept**_ in `historyDoc` when rendering the diff between 2 snapshots.


dmonad:
Thank you! 
I'm going to publish this in a sec.

PR #181, opened 2025-04-11
fix: use TextSelection.between for restoring the selection

nperez0111:
This PR updates the `restoreRelativeSelection` method to use [`TextSelection.between`](https://prosemirror.net/docs/ref/#state.TextSelection%5Ebetween) instead of [`TextSelection.create`](https://prosemirror.net/docs/ref/#state.TextSelection%5Ecreate) when restoring a text selection. This resolves an issue where the text selection may be pointing "between" nodes which are invalid to have text selections into. The `TextSelection.between` method will account for this and try to restore the selection into an appropriate position.

Prosemirror has it's own concept of a relative position (via [bookmarks](https://prosemirror.net/docs/ref/#state.SelectionBookmark)) which can be mapped across transactions. And, the mapping can sometimes [produce invalid positions (relative to the selection type)](https://prosemirror.net/docs/ref/#state.SelectionBookmark.resolve) which is how bookmarks are normally resolved back to selections:
https://github.com/ProseMirror/prosemirror-state/blob/400e561e65cac59968666b57087e3abae3185061/src/selection.ts#L316

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -4,7 +4,7 @@
 
 import { createMutex } from 'lib0/mutex'
 import * as PModel from 'prosemirror-model'
-import { AllSelection, Plugin, TextSelection } from "prosemirror-state"; // eslint-disable-line
+import { AllSelection, Plugin, TextSelection, NodeSelection } from "prosemirror-state"; // eslint-disable-line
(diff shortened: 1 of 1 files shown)


dmonad:
Thank you. This is great! 
I'll make a release in a moment.

PR #182, opened 2025-04-15
refactor: use selection bookmarks to support all possible selection types

nperez0111:
This completely overhauls selection restoration to be based on [Prosemirror Bookmarks](https://prosemirror.net/docs/ref/#state.SelectionBookmark) which are used to store & restore selections independent of the current document.
This is useful for y-prosemirror, since it means that we do not have to implement each selection type, yet all of them will work.

These also fallback to a [TextSelection.between](https://prosemirror.net/docs/ref/#state.SelectionBookmark.resolve) if a bookmark, if invalid or unimplemented, which is a good default.


tommoor:
This would be great – nice work!

PR #184, opened 2025-04-23
Add escape hatch to handle node creation errors

mayacoda:
This PR introduces an additional option `handleNodeCreationFailure` to the sync plugin to call when the conversion between a Y.js type → ProseMirror node fails. If the function returns true, meaning it handles the failure, the type is not deleted. Otherwise, if it's not defined or returns false, the type is deleted.

The reasoning behind this change is to solve mismatched client schemas, exactly as outlined [in this discussion](https://discuss.yjs.dev/t/handling-unknown-nodes-marks-in-prosemirror-schema/3683). 

While schema versioning the documents and editors themselves would work, this approach would provide a nicer user experience (erroring only when an unknown node type is introduced, prompting the user to update their client), and will be less error-prone in later development (no need to maintain schema versions, designate changes as breaking/non-breaking, ensure version checks in all the right places).

Usage:

```javascript
const prosemirrorView = new EditorView(document.querySelector('#editor'), {
  state: EditorState.create({
    doc,
    schema,
    plugins: [
      ySyncPlugin(type, {
        mapping, handleNodeCreationFailure: (node, err) => {
          if (err.message.includes("Unknown node type")) {
            // likely schema mismatch
            // lock down editor, kick out user, etc.
            return true;
          }
          // might be malformed node due to concurrent editing, let y-prosemirror clean up
          return false;
        }
      }),
      ...
    ].concat(exampleSetup({schema}))
  })
})
```

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -66,6 +66,7 @@ export const isVisible = (item, snapshot) =>
  * @property {Y.PermanentUserData|null} [YSyncOpts.permanentUserData]
  * @property {ProsemirrorMapping} [YSyncOpts.mapping]
  * @property {function} [YSyncOpts.onFirstRender] Fired when the content from Yjs is initially rendered to ProseMirror
+ * @property {function(any, Error): boolean | null} [YSyncOpts.handleNodeCreationError] Called when a node cannot be created from a Yjs type. Return false to continue with default behavior (removing the node), true to handle the error.
  */
 
 /**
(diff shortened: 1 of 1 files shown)


dmonad:
Hi @mayacoda,

Thanks for the PR. But I'm currently not accepting any quick-fixes for y-prosemirror. This is not relevant for most users. I also think that there is a better solution for this: There should be an "unknown node" for schema mismatches.

Yjs and all editor bindings are financed by sponsors. If your company wants my support, I'd be seriously happy to work with you. But otherwise, I'll not be working on issues that don't concern me or my users.

artman:
> Yjs and all editor bindings are financed by sponsors. If your company wants my support, I'd be seriously happy to work with you. But otherwise, I'll not be working on issues that don't concern me or my users.

Happy to sponsor this work, will reach out.

PR #185, opened 2025-04-24
fix: clamp positions if out of bounds

nperez0111:
On looking into this comment: https://github.com/yjs/y-prosemirror/pull/176#issuecomment-2825364764 it turns out that he is right, I was able to get a failed selection restoration because the position was restored outside the bounds of the document. I'm not entirely clear how this occurs, but clamping the position to the document size seems like the right thing to do

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -272,7 +272,10 @@ const restoreRelativeSelection = (tr, relSel, binding) => {
         binding.mapping
       )
       if (anchor !== null && head !== null) {
-        tr.setSelection(TextSelection.between(tr.doc.resolve(anchor), tr.doc.resolve(head)))
+        const maxsize = tr.doc.content.size - 1
+        const clampedAnchor = math.min(math.max(anchor, 0), maxsize)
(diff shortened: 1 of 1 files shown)


dmonad:
Wouldn't this suggest that there is a deeper underlying issue?

Just setting the cursor to the end of the document doesn't seem like the right solution. :/

nperez0111:
> Wouldn't this suggest that there is a deeper underlying issue?
> 
> Just setting the cursor to the end of the document doesn't seem like the right solution. :/

Totally fair, my current understanding is that the relative selection can point at a node which was "pushed down" by new content that came in. Using his example [here](https://stackblitz.com/edit/vitejs-vite-jxfgzmld), I reproduced this:

image

So, the relative selection when resolved back to an absolute position is a much larger value, I'm unclear on why that is though 🤔. One idea is that the document at that point in time has both the new & old content, even though the old content should be considered deleted and not count towards the document size it is?

Interestingly, if using my other PR, this issue doesn't seem to occur: https://github.com/yjs/y-prosemirror/pull/182

PR #193, opened 2025-07-16
Fix undo inside custom element

ggodlewski:
While using prosemirror inside Custom Element this line fails.
In such case `this.prosemirrorView._root` is set to `ShadowRoot` not `Document`.
Method createRange exists only in `Document`.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -372,7 +372,7 @@ export class ProsemirrorBinding {
 
     if (selection == null || selection.anchorNode == null) return false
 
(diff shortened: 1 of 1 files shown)


dmonad:
If you use `dom.doc`, you are creating a range from the custom element, which use a different range constructor. It is not immediately obvious to me why this would work. I feel that `prosemirrorView._root` (using the document from the custom element) is the appropriate choice here.

So I assume that your fix simply disables this feature. Could you please verify?

The issue is probably more complex. I'd like to have a reproducible example first and understand why the solution works appropriately in all cases. Please make sure to test that the PR actually works for custom elements.

PR #196, opened 2025-09-14
fix: Prosemirror inside Webcomponent needs to use ownerDocument

serranoio:
When Prosemirror is initialized inside a webcomponent, `this.prosemirrorView._root` is a `ShadowRoot` which does not have a`createRange` method available on it, causing this error.

Screenshot 2025-09-14 at 15 37 10

* This is backwards compatible:
If `createRange` function is available, `_root` is `Document`, so no change
if not, `_root` we assume it is `ShadowRoot`, and use the `.ownerDocument` function. 

* the code does not error on `getSelection`, but for consistency, I made the change here too 👍 

Lmk if you have any questions :). I'd be happy to repro as well

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -373,7 +373,6 @@ export class ProsemirrorBinding {
 
     if (selection == null || selection.anchorNode == null) return false
 
-    
     const range = document.createRange()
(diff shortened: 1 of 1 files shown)


serranoio:
Just saw https://github.com/yjs/y-prosemirror/pull/193. This is the same issue. However, this solution is backwards compatible

PR #198, opened 2025-10-15
fix: compare mark names in equalYTextPText

### LINKED ISSUE #197: equalYTextPText does not compare mark names, only attributes
**Describe the bug**
sync-plugin does not detect changes when mark changed. Based on what I see in code, I assume that the problem is in `equalYTextPText` function, it compares attributes, but does not actually check mark types.

**To Reproduce**
1. Type some text and mark it with `bold`
2. Change mark to `code`. I'm using tiptap, so I run `toggleCode` command that unsets `bold` mark and sets `code` mark.
3. The change won't be applied to the ydoc/not sent to the server

**Expected behavior**
The plugin is to able to recognize mark change.

aonnikov:
Closes #197 

When comparing text nodes `equalYTextPText` used to compare number of marks and their attributes, but not names. This did not work properly for marks without attributes. When changed one mark without attributes to mark with different type without attributes, the function result was negative.

Accompanied the fix with unit test that demonstrates the issue.

### DIFF
--- a/src/plugins/sync-plugin.js
+++ b/src/plugins/sync-plugin.js
@@ -964,7 +964,8 @@ const equalYTextPText = (ytext, ptexts) => {
       object.every(d.attributes, (attr, yattrname) => {
         const markname = yattr2markname(yattrname)
         const pmarks = ptexts[i].marks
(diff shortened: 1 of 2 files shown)


dmonad:
We are working on a full rewrite. Thank you for the PR @aonnikov ! But I simply don't have the time to review all PRs & issues for the previous release before we move on. I hope you understand.

PR #203, opened 2025-11-15
feat: delta bindings

nperez0111:
This is still an in-progress PR, but I'm posting to show an issue that I'm running into.

Specifically with this section of code:

```js
      let deletedSize = 0
      deletedNodes.content.forEach(n => {
        if (n.isText) {
          deletedSize += n.text.length
        } else {
          // TODO this is difficult to get right, since we can't get the currentOp's parent delta
          // I need to delete the next sibling of the parent node, ideally I'd have a wat to traverse up the delta tree to get the parent delta
          deletedSize += 1
        }
      })
      currentOp.delete(deletedSize)
```

What I'm trying to do is take the ProseMirror `Slice`, and itrate through it to build up the delta which would delete this section of content. What I'm running into though is that I've already created a delta which has iterated to the correct position using delta's counting (to create the correct retains using the `deltaPathToDelta` utility).

What would be ideal in this situation is a way to "close" the current delta, and be placed into the parent delta, so that I can continue deleting the slice's content. But, there is likely another approach that I could be doing.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -69,7 +69,8 @@ export class YEditorView extends EditorView {
     const initialPDelta = pstateToDelta(this.state)
     console.log('initialPDelta', initialPDelta.isDone)
     const d = ytype.getContent(Y.noAttributionsManager, { deep: true })
-    console.log(JSON.stringify(d.toJSON(), null, 2))
+    // TODO getContent returns a done delta, so I need to slice it to create a new one & rebase it to the initialPDelta
+    // TODO ideally their would be a `.clone()` method for this. Also unclear why it needs to be marked as done?
     const initialYDelta = /** @type {ProsemirrorDelta} */ (d).slice(0, d.childCnt).rebase(initialPDelta, true)
     this.y.applyDelta(initialPDelta)
     this.dispatch(deltaToPSteps(this.state.tr, initialYDelta))


dmonad:
Constructing the delta in one go is probably going to be hard - especially if you want to compute changes for several changes.

I would traverse from the parent and compute the delta recursively, instead of getting a delta that points to the middle of an existing delta.

```js
// rough outline:
const stepToDelta = (step, beforeDoc, afterDoc, path, currPathDepth, currentDelta = delta.create()) => {
  // iterate through the children of path.node(currPathDepth) and apply the changes to currentDelta
  ..
  // if the path points into a child of the current parent, call stepToDelta recursively
  stepToDelta(step, beforeDoc, afterDoc, path, currPathDepth + 1, currentDelta)
}
```

But I was thinking to simply diff the common parent, which requires less logic for individual edge-cases in prosemirror. The only drawback is that we would need to create a full delta representation of the common parent. But this can be heavily optimized by caching delta⇐⇒prosemirror-node mappings. Something like this could work:

```js
// contains the latest representation of the prosemirror doc in the delta format
const currentDeltaState = delta.create()

// Assume we want to compute the changes for a single `step`

const startPos = oldDoc.resolve(step.from)
const endPos = oldDoc.resolve(step.to)
const sharedDepth = startPos.sharedDepth(endPos)
const commonParentOld = startPos.node(sharedDepth)
// parent offset didn't change, no need to map the position
const commonParentNew = newDoc.nodeAt(startPos.start(sharedDepth))

// the current delta representation of commonParentOld
let deltaParent = currentDeltaState
for (let i = 0; i < sharedDepth; i++) {
  deltaParent = deltaParent.get(startPos.index(i))
}
// delta representation of the new content
// @todo the creation of this delta can be optimized by reusing previous references
const newDeltaParent = pmNodeToDelta(commonParentNew)
const diff = deltaParent.diff(newDeltaParent)
```

I don't believe there can be a good ProseMirror binding without any diffing anyway (unless you have full knowledge of the schema). When we apply changes from Yjs to ProseMirror, ProseMirror may apply additional implicit changes (e.g. delete a node, because the schema is invalid, insert an additional paragraph). We need to figure out these implicit changes and apply them back to Yjs (or cache them so we can rebase future operations). When applying changes to ProseMirror, we need to do something like this, to figure out the implicit changes:

```js
const ydocChanged = eventDelta => {
  const tr = applyDeltaToProsemirror(pmdoc, eventDelta)
  const actuallyAppliedChanges = computeDeltaFromProsemirrorTransaction(tr)
  // compute the difference between expected and actual changes
  const implicitChanges = eventDelta.diff(actuallyAppliedChanges)
  ytype.applyDelta(implicitChanges)
}
```

dmonad:
Thanks @nperez0111 ! 
I continued work using your branch. I hope to finish the binding this weekend and add attribution support.

PR #228, opened 2026-04-12
fix: improve cursor updates in wrapped editors

samwillis:
Hey 👋

I've been experimenting with the new version of this integration from `main` with yjs v14. *Really* excited what this is going to bring!

The demo I have been building that I'm hoping to port to this uses [react-prosemirror](https://github.com/handlewithcarecollective/react-prosemirror), and it seems the cursor-plugin isn't quite compatible with it. The symptoms were that the remote cursor would only appear on first render and then not update in real time.

I worked with Cursor to debug it, and this is what it came up with, and it seems fairly sure it's due to the way react-prosemirror wraps the editor view with its own render layer. The changes seem logical to me.

This is its description of its changes:

---

We changed the cursor plugin to make collaborative cursor updates more reliable in wrapped ProseMirror editors.

What changed:
- redraw cursor decorations on incoming awareness changes by dispatching the plugin transaction
- stop relying only on `view.hasFocus()` and instead treat the editor as focused when the active DOM element is the editor or one of its descendants
- publish cursor updates from real selection movement signals like `selectionchange`, `keyup`, `mouseup`, and `touchend`
- coalesce those updates in a microtask so they don’t fire excessively

Why:
- cursor state could appear on load but not update live as another user moved their caret or edited
- in practice, updates were sometimes only reflected after a reload, which suggested the plugin was missing normal selection/focus timing in our integration

Suspicion:
- this looks related to how `@handlewithcare/react-prosemirror` wraps the editor view
- the stock cursor plugin logic seems to assume more direct `EditorView` focus/update behavior than we were seeing, so these changes make the plugin less dependent on those assumptions

### DIFF
--- a/src/cursor-plugin.js
+++ b/src/cursor-plugin.js
@@ -180,9 +180,11 @@ export const yCursorPlugin = (
       },
       apply (tr, prevState, _oldState, newState) {
         const ySyncMeta = $syncPluginStateUpdate.nullable.expect(tr.getMeta(ySyncPluginKey) || null)
+        const ySyncTransaction = tr.getMeta('y-sync-transaction')
         const yCursorState = tr.getMeta(yCursorPluginKey)
         if (
           (ySyncMeta) ||
+          (ySyncTransaction) ||
           (yCursorState && yCursorState.awarenessUpdated)
         ) {
           return createDecorations(
--- a/src/cursor-plugin.js
+++ b/src/cursor-plugin.js
@@ -203,11 +203,29 @@ export const yCursorPlugin = (
       }
     },
     view: (view) => {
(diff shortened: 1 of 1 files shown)


nperez0111:
Thanks @samwillis, I'll review this

nperez0111:
Hi @samwillis to get back to you on this, this took a bit more of a complicated approach than I had liked so I went with a different direction in https://github.com/yjs/y-prosemirror/pull/229

I did test it out with react-prosemirror & the new approach is compatible with that too, so I think we should be good here once that is merged in

dmonad:
Let's close this PR for now. I opened a ticket to track react-prosemirror compatibility: https://github.com/yjs/y-prosemirror/issues/234

PR #229, opened 2026-04-20
feat: add cursorTimeout option and y-sync-transaction decoration rebuild

nperez0111:
## Summary

- Adds a configurable `cursorTimeout` option (default `10000` ms) to `yCursorPlugin` that controls how long after the editor loses focus before the cursor is cleared from awareness, replacing the immediate-clear-on-blur behavior
- Adds `y-sync-transaction` meta check in `state.apply` so remote Y.js document changes trigger a full decoration rebuild from relative positions instead of just remapping

## Motivation

The current cursor plugin has two issues:

**1. Aggressive cursor clearing on blur causes awareness churn**

The existing behavior immediately sets the cursor to `null` in awareness on every `focusout` event and on plugin `destroy`. This creates unnecessary awareness broadcasts in several scenarios:
- Transient focus loss (clicking a toolbar, dialog, etc.) causes the cursor to disappear and reappear
- Framework lifecycle effects (e.g., React StrictMode double-mounting) destroy and recreate plugins, causing rapid null/set/null/set churn
- `focusout` events are unreliable and may not always fire, leading to inconsistent state

The new approach replaces immediate clearing with a configurable stale timeout. When the editor loses focus, a timer starts. If the user doesn't return within `cursorTimeout` ms, the cursor is cleared. This reduces awareness broadcasts while still allowing cursors to expire.

| `cursorTimeout` value | Behavior |
|---|---|
| `> 0` (default `10000`) | Clear cursor after this many ms without focus |
| `null` or `0` | Never auto-clear; cursor persists until awareness drops |

**2. Missing decoration rebuild on remote document changes**

The cursor plugin's `state.apply` checked for `ySyncPluginKey` meta (user reconfiguration commands) but not `y-sync-transaction` meta (remote Y.js changes arriving via the sync plugin). These are set on different transactions and never overlap. Without the `y-sync-transaction` check, remote document changes only remap existing decorations through the transaction mapping rather than rebuilding them from relative positions, which can lead to stale or incorrect cursor positions after remote edits. The undo plugin already checks for both metas.

### `src/cursor-plugin.js`

- Added `cursorTimeout` option (default `10000`) to `yCursorPlugin` signature
- Added `tr.getMeta('y-sync-transaction')` as a third condition for full decoration rebuild in `state.apply`
- Added `resetStaleTimer()` mechanism in `view()` closure that starts/resets a timeout when the cursor is published
- Modified `updateCursorInfo`: when focused, publishes cursor and resets the stale timer; removed the else-if branch that immediately cleared cursor on blur
- Removed `focusout` event listener entirely
- Modified `destroy`: clears the stale timer and removes listeners, but does NOT clear cursor from awareness (lets the awareness protocol handle expiry naturally)

### `tests/cursor.test.js` (new, 11 tests)

- `testCursorPublishedOnFocus` -- cursor appears in awareness when editor is focused
- `testCursorNotClearedImmediatelyOnBlur` -- cursor persists after blur (no immediate clearing)
- `testStaleTimerClearsCursorAfterTimeout` -- cursor is cleared after `cursorTimeout` ms when blurred
- `testCursorTimeoutNullDisablesStaleClearing` -- `null` disables auto-clearing
- `testCursorTimeoutZeroDisablesStaleClearing` -- `0` disables auto-clearing
- `testRefocusResetsStaleTimer` -- re-focusing before timeout resets the timer
- `testDestroyDoesNotClearCursor` -- destroying plugin does not clear cursor from awareness
- `testDestroyCleanupStaleTimer` -- destroying plugin cancels pending stale timer
- `testRemoteAwarenessUpdatesRebuildDecorations` -- awareness changes trigger decoration rebuild
- `testYSyncTransactionTriggersDecorationRebuild` -- `y-sync-transaction` meta triggers rebuild
- `testCursorUpdatesOnSelectionChange` -- selection changes update cursor in awareness

### Test runner updates

- `tests/index.js` and `tests/index.node.js` import and register the new `cursor` test module

## Related

Partially addresses #228 (the `y-sync-transaction` decoration rebuild fix). Does not include the other changes from that PR (DOM event listeners, `isEditorFocused`, `scheduleCursorUpdate`) as those target react-prosemirror-specific wrapping issues.

### DIFF
--- a/src/cursor-plugin.js
+++ b/src/cursor-plugin.js
@@ -91,7 +91,7 @@ export const createDecorations = (
    */
   const decorations = []
   awareness.getStates().forEach((aw, clientId) => {
-    if (!awarenessFilter(doc.clientID, clientId, aw)) {
+    if (!awarenessFilter(awareness.clientID, clientId, aw)) {
       return
     }
 
--- a/src/cursor-plugin.js
+++ b/src/cursor-plugin.js
@@ -152,6 +152,7 @@ export const createDecorations = (
  * @param {(user: User, clientId: number) => HTMLElement} [opts.cursorBuilder] A function that creates a cursor element
  * @param {(user: User, clientId: number) => import('prosemirror-view').DecorationAttrs} [opts.selectionBuilder] A function that creates a selection decoration
  * @param {(state: import('prosemirror-state').EditorState) => {$anchor: import('prosemirror-model').ResolvedPos, $head: import('prosemirror-model').ResolvedPos}} [opts.getSelection] A function that gets the selection from the editor state
+ * @param {number | null} [opts.cursorTimeout] How long (in ms) after the editor loses focus before the cursor is cleared from awareness. Set to `null` or `0` to disable (cursor persists until awareness drops). Defaults to `10000`.
  * @param {string} [cursorStateField] By default all editor bindings use the awareness 'cursor' field to propagate cursor information, this allows you to use a different field name
  * @return {any}
  */
@@ -161,7 +162,8 @@ export const yCursorPlugin = (
     awarenessStateFilter = (currentClientId, userClientId) => currentClientId !== userClientId,
     cursorBuilder = defaultCursorBuilder,
     selectionBuilder = defaultSelectionBuilder,
-    getSelection = (state) => state.selection
+    getSelection = (state) => state.selection,
+    cursorTimeout = 10000
   } = {},
   cursorStateField = 'cursor'
 ) =>
@@ -180,9 +182,11 @@ export const yCursorPlugin = (
       },
       apply (tr, prevState, _oldState, newState) {
         const ySyncMeta = $syncPluginStateUpdate.nullable.expect(tr.getMeta(ySyncPluginKey) || null)
+        const ySyncTransaction = tr.getMeta('y-sync-transaction')
         const yCursorState = tr.getMeta(yCursorPluginKey)
         if (
           (ySyncMeta) ||
+          (ySyncTransaction) ||
           (yCursorState && yCursorState.awarenessUpdated)
         ) {
           return createDecorations(
@@ -208,6 +212,29 @@ export const yCursorPlugin = (
         if (view.docView) { // TODO why is this using docView? Ask Kevin about this.
           view.dispatch(view.state.tr.setMeta(yCursorPluginKey, { awarenessUpdated: true }))
         }
+      }
+      /**
+       * @type {ReturnType<typeof setTimeout> | null}
+       */
+      let staleTimer = null
+      const resetStaleTimer = () => {
+        if (staleTimer != null) {
+          clearTimeout(staleTimer)
+          staleTimer = null
+        }
+        if (cursorTimeout == null || cursorTimeout <= 0) return
+        staleTimer = setTimeout(() => {
+          staleTimer = null
+          if (!view.hasFocus()) {
+            const current = awareness.getLocalState() || {}
+            if (current[cursorStateField] != null) {
+              awareness.setLocalStateField(cursorStateField, null)
+            }
+          } else {
+            // still focused, re-arm the timer
+            resetStaleTimer()
+          }
+        }, cursorTimeout)
       }
       const updateCursorInfo = () => {
         const ystate = ySyncPluginKey.getState(view.state)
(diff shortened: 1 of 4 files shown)


dmonad:
The main motivation for the timeout seems to be that you want to prevent remote cursors disappearing when you select things in a different menu (that may be part of the editor).

But a "lingering awareness state" seems like a bad idea as well. When I click somewhere else, I expect that my cursor disappears for all users.

I'm sure there are better solutions to this problem.

1. We could expand the focus-area to include the parent of the menu item.
2. We could add additional conditionals (e.g. cursor must leave the focusarea AND click somewhere else - remove awareness only when a click outside the focusarea happens).

Focus might have been the wrong choice to begin with, because you might lose it for weird reasons. 

Let's keep discussing how to properly solve this issue.

nperez0111:
Happy to discuss @dmonad:

So, I went this route because:
 1. focus moving to another element is _very hard to track_, as well as being imprecise. People have all sorts of patterns for menu items, and a shared parent is not enough for menus which have been brought up to the document.body (which is a common solution to z-index issues). So, tracking by element focus is not a solution in my opinion.
  2. I'd like to challenge this assumption: "When I click somewhere else, I expect that my cursor disappears for all users." why would you expect this? Say you clicked to another browser tab, to copy & paste something into the editor. Would you really _need_ for the other user to know that you left the tab and came back? What purpose does this serve? In my experience, I would prefer for the cursor to linger a little longer than to flicker in & out. Even so, I believe this should be an application concern, rather than our library, which is why I made the timeout configurable. If you wanted your behavior, you can set the timeout to 0, and it'd disappear instantly. I find the cursor flickering in & out to be more distracting, so I'd prefer to avoid that if possible, I believe a timeout to be simplest in resolving this than any number of heuristics we try to come up with.
  3. It already is acceptable to have a "stale cursor", like for example a user closes their browser tab, we hold their awareness state for however long it is configured for (e.g. 30s). This is acceptable because there isn't really a great way to do this otherwise as well as the reality being that a user cursor isn't really much in the way of anything, it can only be a minor annoyance for so long, but it really isn't in the way of anything.

dmonad:
>     1. focus moving to another element is _very hard to track_, as well as being imprecise. People have all sorts of patterns for menu items, and a shared parent is not enough for menus which have been brought up to the document.body (which is a common solution to z-index issues). So, tracking by element focus is not a solution in my opinion.

I'm not arguing about that. But a timeout would lead to weird behavior, especially when combined with other bindings. I just wonder if there are better solutions. I don't understand the problem in detail. A default 10s lingering default would be inconsistent with all other editor bindings that I have created.

>     2. I'd like to challenge this assumption: "When I click somewhere else, I expect that my cursor disappears for all users." why would you expect this? Say you clicked to another browser tab, to copy & paste something into the editor. Would you really _need_ for the other user to know that you left the tab and came back? What purpose does this serve? In my experience, I would prefer for the cursor to linger a little longer than to flicker in & out. Even so, I believe this should be an application concern, rather than our library, which is why I made the timeout configurable. If you wanted your behavior, you can set the timeout to 0, and it'd disappear instantly. I find the cursor flickering in & out to be more distracting, so I'd prefer to avoid that if possible, I believe a timeout to be simplest in resolving this than any number of heuristics we try to come up with.

When I click in another prosemirror editor, I expect the cursor from the original editor to disappear. You can't be in two places at once.

>     3. It already is acceptable to have a "stale cursor", like for example a user closes their browser tab, we hold their awareness state for however long it is configured for (e.g. 30s).

Actually, y-websocket attempts to clear their awareness state on tab close, but that does not always succeed. The backend is supposed to clear the awareness state on disconnect. y-websocket-server implements this behavior correctly. If this does not work in yhub, it would be a bug.

---

Another option would be to disable clearing awareness on focuschange. Then the editor binding (or the user) would be responsible for figuring out when awareness changed.

By "click outside the focusarea" I mean clicking on an element that is not anchored inside the editor dom element + additional checks if applicable.

Personally, I like the default behavior to clear awareness on "click outside the focusarea". This only leads to weird issues when working with weird menus that use fixed position and are anchored outside the parent element.

dmonad:
I just noticed that this would destroy awareness state of other bindings (other bindings may reuse the awareness state). So lingering timeout delete is an absolute no-go.

dmonad:
We have to be aware of the fact that others use this binding in weird ways. I don't like to add too many interactions. If there are interactions, they should be immediate - otherwise this binding will mess with other bindings. I recommend that we simply add an option to disable clearing awareness.

For Blocknote, I recommend that you also consider that users might have two editors side-by-side. it would be weird to see one user active in two documents at the same time. Also, too many remote cursors make it hard to read the content.

PR #230, opened 2026-04-21
test: appendTransaction and filterTransaction plugin interactions with sync

nperez0111:
Here are the bunch of test cases I was able to add in. It's nice to see that things are still working with these sorts of more complex plugins. So, let's keep it that way with these tests.

### AI Summary
---

Adds tests in `tests/delta.test.js` covering how y-prosemirror syncs changes produced by common ProseMirror plugin patterns across two clients:

- `appendTransaction` assigning node attributes (unique block IDs via `uuidv4()`), including after inserting new paragraphs
- Ephemeral `state.apply()` calls (the pattern that caused [prosemirror-inputrules#2](https://github.com/handlewithcarecollective/prosemirror-inputrules/pull/3)) don't leak into synced state
- `filterTransaction` plugins that allow sync transactions through while gating user edits (read-only client)
- `appendTransaction` adding marks (auto-link URL detection)
- Chained `appendTransaction` plugins where one reads another's output

### DIFF
--- a/tests/delta.test.js
+++ b/tests/delta.test.js
@@ -2,11 +2,12 @@ import * as t from 'lib0/testing'
 import * as YPM from '@y/prosemirror'
 import * as basicSchema from 'prosemirror-schema-basic'
 import * as Y from '@y/y'
-import { EditorState } from 'prosemirror-state'
+import { EditorState, Plugin, PluginKey } from 'prosemirror-state'
 import { Fragment, Schema, Slice } from 'prosemirror-model'
 import * as delta from 'lib0/delta'
 import { findWrapping, ReplaceAroundStep } from 'prosemirror-transform'
 import { EditorView } from 'prosemirror-view'
+import { uuidv4 } from 'lib0/random'
 
 const schema = new Schema({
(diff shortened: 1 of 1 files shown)


dmonad:
Tests that break would help a lot. But the tests seem to work fine. 

More is not always better. The diff to Yjs is calculated after changes have been dispatched. So any test that checks "ephemeral state" or `appendTransaction` is superfluous and adds unnecessary maintenance burden to the project. AI produces cheap code and does not help in projects where we have to work out the details.

nperez0111:
@dmonad while I would agree that these specific tests are "cheap", and are a likely a bit of noise. I would push back on "any test that checks 'ephemeral state' or appendTransaction is superfluous".

The reason I added these tests is that a prosemirror binding does _need_ to be able to pass them (otherwise they would not work with other plugins). So, whether the current implementation passes the tests or not is sort of irrelevant. The tests are valid, but they definitely can be more minimal and pointed, I accept that. If we decided to change the binding tmrw, and these tests did not pass, that would be valuable to know, I think.

PR #231, opened 2026-04-21
fix: implicitly create "filled" prosemirror nodes to not crash the editor

nperez0111:
Hi @dmonad, to fix the blocknote demo (which is currently broken), I was going to apply the same fix that [I did here](https://github.com/yjs/y-prosemirror/pull/225#discussion_r3051464129).

After our recent fitting conversation, it made me realize that we can instead fix this by changing the nodes that y-prosemirror generates. 
One thing that is sort of nice about this solution, is that every editor would deterministically fill in the same way (assuming they have the same schema).
So, it seems to just work? Maybe you would know better what could go wrong with this, but from my understanding this is a fine solution to have (and unblocks me by allowing me to use blocknote again).

Let me know what you think!

### DIFF
--- a/src/sync-utils.js
+++ b/src/sync-utils.js
@@ -325,9 +325,19 @@ export const deltaToPNode = (d, schema, dformat) => {
     attrs[attr.key] = attr.value
   }
   const dc = d.children.map(c => delta.$insertOp.check(c) ? c.insert.map(cn => deltaToPNode(cn, schema, c.format)) : (delta.$textOp.check(c) ? [schema.text(c.insert, formattingAttributesToMarks(c.format, schema))] : []))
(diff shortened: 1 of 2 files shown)


dmonad:
I think this is a fine solution for now. 

Something to note: createAndFill will create additional nodes on all clients. In the case of blockquote, every client will create the missing paragraph. However, this change will only be picked up by yjs on the next keystroke (when the diff picks up the implicitly added changes), so there is less chance of content duplication.

The ideal solution would notice this as an implicit change (e.g. by rebasing diffs) and not apply it to the Yjs structure. Alternatively, we go for something similar to y-prosemirror v1, which does not recognize the changes until the next change to that specific node. 

A note on the tests: I love them! But they should be meaningful and not add unnecessary noise. Can you please remove the edits to the test file before we merge?

nperez0111 (commented):
I think these tests are better at showing what I was going for. I believe it to be an important feature of our sync system (blocknote breaks without it), so I won't remove it altogether.

PR #232, opened 2026-04-22
test: add failing repros for lib0 delta diff + suggestion reconfigure crash

nperez0111:
## Summary

Adds two failing test cases captured while working on a BlockNote-style suggestion demo. Both reproduce bugs I'd like to see fixed upstream — opening the PR with the repros so a fix can be landed against them.

### 1. `testLib0DeltaDiffUnexpectedCaseRepro`

Tight repro of an `Unexpected case` throw from `lib0/delta`'s `diff()` when given two structurally asymmetric deltas. The pair was captured at `src/sync-plugin.js:339` in a running demo: on the subsequent `update()` cycle after `configureYProsemirror` round-tripped the PM view between (suggestionDoc + attributionManager) and (mainDoc, no attribution manager), the Y-side delta carried a suggestion-attributed text leaf **and** a top-level `delete 1` sibling, while the PM-side delta had no text child in the corresponding paragraph and no root-level delete. `diff2` / `applyChangesetToDelta` in `lib0/src/delta/delta.js` cannot reconcile the asymmetry and throws.

### 2. `testReconfigureRoundTripWithSuggestions`

End-to-end repro of the same underlying crash surfaced through `y-prosemirror`. Sequence:

1. Bind view to suggestionDoc + AM
2. Insert suggestion-mode content (renders with insertion marks)
3. Reconfigure to main doc (no AM) — strips marks
4. Reconfigure back to suggestionDoc + AM — crash
5. Type anything on the rebound view — crash (via the sync-plugin update)

Both tests fail today and should pass once the underlying issue is fixed.

## Test plan
- [x] `npm test` — both tests currently fail:
  - `[39/80] suggestions: lib0 delta diff unexpected case repro — Failure`
  - `[42/80] suggestions: reconfigure round trip with suggestions — Failure`
- [ ] Reviewer: decide whether this needs a matching fix PR against `lib0` or can be addressed inside `src/sync-plugin.js`

### DIFF
--- a/tests/suggestions.test.js
+++ b/tests/suggestions.test.js
@@ -774,6 +774,123 @@ export const testReconfigureAfterDeletion = async () => {
   })
 }
 
+/**
+ * Tight repro of a `Unexpected case` throw from `lib0/delta`'s `diff()` when
+ * given two structurally asymmetric deltas.
+ *
+ * The pair was captured at `src/sync-plugin.js:339` in a running BlockNote
+ * demo: on the subsequent `update()` cycle after `configureYProsemirror`
+ * round-tripped the PM view between (suggestionDoc + attributionManager) and
+ * (mainDoc, no attribution manager), the Y-side delta carried a suggestion-
+ * attributed text leaf AND a top-level `delete 1` sibling while the PM-side
+ * delta had no text child in the corresponding paragraph and no root-level
+ * delete. `diff2` / `applyChangesetToDelta` in `lib0/src/delta/delta.js`
+ * cannot reconcile the asymmetry and throws.
+ *
+ * Test fails today (throws); flip the assertion to `!t.fails` once the
+ * upstream fix in lib0 lands.
+ */
+export const testLib0DeltaDiffUnexpectedCaseRepro = () => {
+  const paraAttrs = {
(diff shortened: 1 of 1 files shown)


dmonad:
Sweet, thanks. I will work on this.

dmonad:
The first test is about diffing deletes (which impossible to implement without knowing what the other delta deletes), and the second one is incorrect. 

It should be:
```js
  const helloWorldBangDoc = {
    type: 'doc',
    content: [
      {
        type: 'paragraph',
        content: [
          { type: 'text', text: 'hello' },
          { type: 'text', text: ' world!', marks: [insertionMark] } // '!' is merged
        ]
      }
    ]
  }
```

If you fix the test, it works correctly

nperez0111:
1. It was an attempt to reproduce with a smaller test case. But, here is the issue in the demo:

https://github.com/user-attachments/assets/4c789059-b9c6-4d2a-9b18-abe2d6b9addf

2. I have to imagine that the issue was actually another mark instance with different attribute values, which does fail: https://github.com/nperez0111/y-prosemirror/commit/7d17166a90bc107055771e1ed948029e516267ef

PR #237, opened 2026-05-11
feat: cursor plugin v2 update

nperez0111:
This takes a different approach from https://github.com/yjs/y-prosemirror/pull/229 by inverting the decision of what the next cursor state should be written into awareness to the consumer (optionally).

By default, the plugin will check whether the editor's view is focused & then set the new cursor state. Otherwise, it will set the cursor to `null` if not focused & the cursor is actually owned by this plugin instance.

This logic encodes the previous behavior, but allows the consumer to override this for more complex integrations (e.g. determining what is focused).

Everything is still handled synchronously. 

There are a couple of additional changes included in the PR:

 - doc synchronization can lag behind cursor synchronization sometimes, so I needed to add a patch for cases where the cursor is ahead and referencing a position that is not yet in the doc content: https://github.com/yjs/y-prosemirror/commit/1b96ad1b3ed3dcb790d801852d92c99cb5069dce
 - Because we rely on `focusin` of the editor, which fires before prosemirror moves the `selection`, there can be a split second where the `focusin` handler syncs the previous selection instead of the selection that would be generated if we were to wait for prosemirror. So I added a condition that will just no-op this event if the `selection` is not exactly the same place where the focus just moved to (allowing prosemirror to move the selection which would re-sync the cursor anyway).
 - Adds tests for cursor synchronization


dmonad (approved):
thx

PR #239, opened 2026-05-12
test: add `appendTransaction` sync tests

nperez0111:
Based on the tests from here: https://github.com/yjs/y-prosemirror/pull/230 I'm adding these minimal test cases because they are the constraints that our sync solution __must__ abide by to work within the ProseMirror ecosystem. Note that this is based on this PR: https://github.com/yjs/y-prosemirror/pull/238

As you can see the ephemeral state test case does fail, because we are applying to the ydoc during `appendTransaction`, which ProseMirror implementations assume to be a pure operation.

### DIFF
--- a/tests/delta.test.js
+++ b/tests/delta.test.js
@@ -2,11 +2,12 @@ import * as t from 'lib0/testing'
 import * as YPM from '@y/prosemirror'
 import * as basicSchema from 'prosemirror-schema-basic'
 import * as Y from '@y/y'
-import { EditorState } from 'prosemirror-state'
+import { EditorState, Plugin } from 'prosemirror-state'
 import { Fragment, Schema, Slice } from 'prosemirror-model'
 import * as delta from 'lib0/delta'
 import { findWrapping, ReplaceAroundStep } from 'prosemirror-transform'
 import { EditorView } from 'prosemirror-view'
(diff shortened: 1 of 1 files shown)


nperez0111:
This case fails because in `appendTransaction`, the sync plugin performing the side-effect of applying to the ydoc, when prosemirror plugins may expect this sort of change to be side-effect free.

To get around this, I had, in the December version, the sync plugin state use a temporary delta to accumulate the prosemirror changes onto & then view.update would actually apply those changes to the ydoc (so that the side-effect was in view.update, not in appendTransaction). But maybe you can come up with another solution. Marijn had suggested just keeping the transactions themselves in the state, which would achieve a similar thing

dmonad:
I thought I could hijack view.dispatch to only update y when state is actually dispatched - but that doesn't work. I removed `appendTransaction` again and reverted to view.update.

There's probably no way around having an async dispatch function. Currently, suggestions and schema-related state-changes are applied synchronously back to the prosemirror state (previously appendTransaction, now view.update handler then view.dispatch). We shouldn't update state within view.update - but I want to keep it synchronous for a bit longer. I planned some changes that remedy the issues of async editor<>yjs synchronization.

I copied one of your test cases - even though we don't update yjs state within appendTransaction any longer.

PR #257, opened 2026-06-01
attributed nodes, required attr, failing test cases

nperez0111:
This PR introduces the required attribute that we talked about on call.

For ProseMirror to not automatically create these attribution nodes, it needs to have a required attribute which will force prosemirror's content fitting to bail when creating intermediate nodes (because the attribute is required but prosemirror has no default to set). This hack allowed us to not mess with prosemirror parsing rules and inject arbitrary nodes into positions of otherwise strict schemas.

This PR introduces some test cases which fail in the same way that BlockNote's schema is failing.

I’ve diagnosed the issue. The problem is that the deltaToTr assumes that a delta’s ops can be mapped 1:1 with a prosemirror step. But, in the case of an attributed node change this turns out to not be true. Let’s see an example:

```html
// Base
<blockContainer>
  <paragraph>abc</paragraph>
</blockContainer>
// Suggestion
<blockContainer>
  <paragraph—attribution>abc</paragraph--atribution>
  <heading>abc</heading>
</blockContainer>
```

But, the delta wants to:

```html
<modify>paragraph -> paragraph--atribution</modify>
<insert>heading>abc</insert>
```

But, in ProseMirror **each step must result in a valid document**, meaning:

* if you do the *modify* first, you may break the schema constraints of blockContainer

* if you do the *insert* first, you have the very issue we are trying to prevent (two nodes where one is allowed)

Some possible solutions are:

* In ProseMirror you can make arbitrary steps for operations you want to perform. We could make a custom ProseMirror step which represents this attribution node replacement in a single step (suggestion attribution step?)

  * This would require reading the delta ahead to make sure that it does an insert directly afterwards. So that you can safely collapse both delta ops into one pm step

* In ProseMirror each step has the abillity to merge with others (e.g. two inserts can be merged to be one longer insert). So, we could accumulate prosemirror steps (maybe a list), and only afterwards, merge the steps together and apply them if still valid. This should merge the two individually invalid steps into 1 valid replacement.

### DIFF
--- a/tests/attributed-nodes.test.js
+++ b/tests/attributed-nodes.test.js
@@ -16,7 +16,7 @@ const schema = new Schema({
     ...nodes,
     'paragraph--attributed': {
       content: 'inline*',
-      group: 'block',
+      group: 'block attributed',
       parseDOM: [{ tag: 'p[data-attributed]' }],
       toDOM () {
         return ['p', { 'data-attributed': 'true' }, 0]
@@ -25,10 +25,19 @@ const schema = new Schema({
     'heading--attributed': {
       attrs: { level: { default: 1 } },
       content: 'inline*',
-      group: 'block',
+      group: 'block attributed',
       defining: true,
       toDOM (node) {
         return ['h' + node.attrs.level, { 'data-attributed': 'true' }, 0]
+      }
+    },
+    container: {
+      attrs: { 'yjs-suggestion-node': { } },
+      content: 'attributed* block attributed*',
+      group: 'block',
+      parseDOM: [{ tag: 'container' }],
+      toDOM () {
+        return ['container', {}, 0]
       }
     }
   },
@@ -56,8 +65,9 @@ const assertDocJSON = (doc, expected, message) => {
  *
  * @param {AttributedNodesPredicate} attributedNodes
  * @param {string} baseContent
+ * @param {import('lib0/delta').Delta} [seedDelta]
  */
-const setup = (attributedNodes, baseContent) => {
+const setup = (attributedNodes, baseContent, seedDelta = delta.create().insert([delta.create('paragraph', {}, baseContent)]).done()) => {
   const doc = new Y.Doc({ gc: false, guid: 'base' })
   const suggestionDoc = new Y.Doc({ isSuggestionDoc: true, gc: false, guid: 'suggestions' })
   const suggestionModeDoc = new Y.Doc({ isSuggestionDoc: true, gc: false, guid: 'suggestions-edit' })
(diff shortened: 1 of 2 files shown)


dmonad:
oh no.. i accidentally merged this.  i will apply the fixes we talked about.

tests/attributed-nodes.test.js
```diff
@@ -88,12 +98,33 @@ const setup = (attributedNodes, baseContent) => {
...
+const canonical = (name) =>
+  name.endsWith('--attributed') ? name.slice(0, -'--attributed'.length) : name
+
+/**
+ * A `container` seed delta holding a single `paragraph` child.
+ * @param {string} text
+ **/
+const containerWithParagraph = (text) =>
+  delta.create().insert([
+    delta.create('container', { 'yjs-suggestion-node': true }, [
```
dmonad, line 123: this is not an attributed node (has no suffix `--attributed`), hence the attribute here is meaningless

tests/attributed-nodes.test.js
```diff
@@ -245,3 +276,226 @@ export const testAcceptFlipsBackToCanonical = _tc => {
...
+    canonical(viewer.state.doc.child(0).child(0).type.name) === 'heading',
+    'after first flip: viewer child is a heading'
+  )
+
+  // heading -> paragraph (flip back)
+  t.assert(
+    canonical(editor.state.doc.child(0).child(0).type.name) === 'heading',
+    'mid-flip: container child is now a heading'
+  )
+  editor.dispatch(editor.state.tr.setNodeMarkup(
```
dmonad, line 413: Creating the reverse operation "flip back" doesn't remove the attribution - this is not expected to work.

tests/attributed-nodes.test.js
```diff
@@ -245,3 +276,226 @@ export const testAcceptFlipsBackToCanonical = _tc => {
...
+  assertDocJSON(editor.state.doc, expected, 'editor: child is canonical heading')
+}
+
+/**
+ * Rejecting a container child-flip suggestion should discard it: every peer
+ * returns to the original canonical `paragraph` child.
+ *
+ * @param {t.TestCase} _tc
+ */
+export const testContainerChildFlipReject = _tc => {
```
dmonad, line 476: Reject doesn't remove the attribution - we talked about this in earlier meetings. This is a separate Yjs issue

tests/attributed-nodes.test.js
```diff
@@ -25,11 +25,20 @@ const schema = new Schema({
...
       content: 'inline*',
-      group: 'block',
+      group: 'block attributed',
       defining: true,
       toDOM (node) {
         return ['h' + node.attrs.level, { 'data-attributed': 'true' }, 0]
       }
+    },
+    container: {
+      attrs: { 'yjs-suggestion-node': { } },
```
dmonad, line 35: The hack was supposed to be that `--attributed` nodes have an attribute without a default. Why did you add it to the container, which is not attributed?

dmonad (commented):
I think we have to clear up our understandings of the "attribution nodes".

An attribution node (suffixed by `--attributed`) can be automatically created by a prosemirror change (e.g. split). So you suggested adding an attribute without a default value to it.

But you added this attribute now to the container, which is not an attribution node and missed it on the actual attribution nodes. A lot of the test cases don't make sense to me at all. My best guess is that this PR was LLM generated and you didn't review it. Or do we have a mismatched understanding of how this should work?

I will salvage the tests that make sense to me and introduce the `y-attribution` attribute to the attribution nodes.

PR #273, opened 2026-08-20
feat: `onInternalError` option to observe recovered sync errors

YousefED:
## Motivation

`YSyncRdt.applyDelta` has a last-resort catch: if the Y-side apply throws mid-transact, it reverts the unappliable part of the change and logs

```
[y/prosemirror] ytype.applyDelta failed - reverting the unappliable part of the change
```

The recovery itself is great — but the *only* signal is a console warning. In BlockNote this let a real bug (an editor-side `removeBlocks` throwing during a re-rendered version diff) stay invisible for two months: the UI silently stopped updating, nothing errored, and no test harness could observe the failure short of scraping the console.

## Change

Adds an `onInternalError?: (err: unknown) => void` option to `YSyncRdt` (threaded through `syncPlugin` opts):

- **Default behavior is unchanged**: `onInternalError` defaults to the existing `console.warn` call, byte-identical message.
- **Providing a handler replaces the warning**, handing the host the original error object (with stack) — so apps can report it to error trackers and test harnesses can fail loudly on it, without double-reporting.

Includes a unit test (`testYSyncRdtOnInternalErrorReplacesWarn`) covering both the handler delivery and the warn suppression. `npm test` (125/125) and `npm run lint` pass.

Happy to adjust naming/placement — and if this direction is right, the same hook could later serve other recovered-error sites (e.g. the cursor-plugin's selection-encode catch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/rdt/y-sync.js
+++ b/src/rdt/y-sync.js
@@ -3,6 +3,15 @@ import * as delta from 'lib0/delta'
 import { $prosemirrorDelta } from '../sync-utils.js'
 
 /**
+ * Default {@link YSyncRdt} `onInternalError` handler: log a console warning
+ * and move on.
+ *
+ * @param {string} context
+ * @param {unknown} err
+ */
+const defaultOnInternalError = (context, err) => console.warn(context, err)
+
+/**
  * The Y side of the sync binding: a thin lib0-`RDT` wrapper around a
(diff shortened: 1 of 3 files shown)


dmonad:
I implemented a very similar API and created a new release

PR #276, opened 2026-09-17
Expose fragmentToTr and pass options through fragmentToPm

nperez0111:
Hey @dmonad so I figured out why again that it was that I needed `fragmentToTr`, we wanted to have a demo where we purposefully don't add a sync plugin binding because we want to diff two documents completely in memory but have y/prosemirror calculate that diff. I was very close to be able to do this with the already exposed `fragmentToPm` but it does not support passing through the 3rd argument object in `fragmentToTr` (i.e. allowing me to use a DiffRenderer). So, I can go either way, on this, either expanding the `fragmentToPm` options to pass through that 3rd argument or just exposing the underlying primitive `fragmentToTr`, I think there is an argument for both anyway honestly, so I just implemented both.


dmonad:
I actually thought I removed this function. It is no longer being used by the plugin. I propose that I expose a function that creates a prosemirror state from a Yjs node. Then you can simply replace the state of the prosemirror document and ysync plugin will automatically pick up the differences. Does that work for you?

nperez0111:
@dmonad, yea that would work for me, so long as it exposes options to include the renderer as well

dmonad:
I'm now exporting `ynodeToPmnode` with all required parameters. It allows configuring the renderer and it uses the same transformer as the ysync plugin - which was broken in the old code.

Issue #278, opened 2026-09-22
Do attribution marks still need excludes: ''?


nperez0111:
@dmonad, do we still need to have `excludes: ''` on the attribution marks? Setting it like that allows multiple instances of the same attribution mark to overlap, and if I switch from one diff preview to another it ends up showing incorrectly because it _should_ be replacing the marks rather than accumulating them.

dmonad:
Now that you mention it, I don't remember at all why we ever set `excludes: ''`. So feel free to try it out so we can update the docs. If switching diff-views does not work, that would be a bug either way.

nperez0111:
Yep, seems to just cause bugs. I'll submit something to change ATTRIBUTION.md for this

PR #279, opened 2026-09-22
Remove excludes guidance from attribution docs

nperez0111:
removes mention of excludes from ATTRIBUTION.md per #278 discussion


dmonad:
thx!

